PR #6, opened 2016-08-09
Report errors via 'error' event.

alexmarkley:
Currently, network errors emitted by the underlying Websockets library are thrown as exceptions instead of emitted as events, because no handler is bound to the 'error' event. Also, because of the execution context at the point where the exception is being thrown, it is effectively uncatchable by the library consumer.

This pull request solves the problem by adding an 'error' event which handles such errors, allowing the library consumer to respond to error conditions appropriately.

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,6 +276,7 @@ The following events can be emitted from the `WebsocketClient`:
 * `open`
 * `message`
 * `close`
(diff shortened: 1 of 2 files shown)


CjS77:
Included in #27

PR #7, opened 2016-08-11
Prevent Pinger Exceptions

alexmarkley:
Under certain error conditions (if the connection is disconnected by the caller during adverse network conditions) the "pinger" setInterval will continue to run, triggering a fatal unhandled exception.

This patch makes gdax-node a little smarter about disabling pinger during any websocket disconnect scenario.


fb55:
@alexmarkley Thanks for this patch — if you're still interested, could you rebase it?

fb55:
Thanks a lot @alexmarkley!

PR #14, opened 2016-11-08
lib/clients/authenticated: Add trailing volume getter

mihar:
Adding support for the new trailing volume endpoint which can be used to estimate the fee rebates.

Documentation is here: https://docs.gdax.com/#user-account


CjS77:
LGTM

PR #15, opened 2016-11-09
Add authentication for websocket feed

CjS77:
This will allow users to receive user-specific messages on the websocket feed in future

### DIFF
--- a/lib/clients/authenticated.js
+++ b/lib/clients/authenticated.js
@@ -1,5 +1,4 @@
 var util = require('util');
-var crypto = require('crypto');
 var querystring = require('querystring');
 var async = require('async');
 var signRequest = require('../../lib/request_signer').signRequest;
--- a/lib/clients/authenticated.js
+++ b/lib/clients/authenticated.js
@@ -2,6 +2,7 @@ var util = require('util');
 var crypto = require('crypto');
 var querystring = require('querystring');
 var async = require('async');
+var signRequest = require('../../lib/request_signer').signRequest;
 
 var _ = {
   'forEach': require('lodash.foreach'),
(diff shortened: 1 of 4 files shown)


lib/clients/authenticated.js
```diff
@@ -2,6 +2,7 @@ var util = require('util');
 var crypto = require('crypto');
```
mihar: Remove dependency if unused

lib/clients/authenticated.js
```diff
@@ -2,6 +2,7 @@ var util = require('util');
...
 var querystring = require('querystring');
```
mihar: Remove dependency if unused

lib/clients/websocket.js
```diff
@@ -2,11 +2,23 @@ var EventEmitter = require('events').EventEmitter;
 var Websocket = require('ws');
 var util = require('util');
 var _  = {assign: require('lodash.assign')};
-
-var WebsocketClient = function(productID, websocketURI) {
+var crypto = require('crypto');
```
mihar: Remove dependency if unused

lib/clients/websocket.js
```diff
@@ -67,6 +85,9 @@ _.assign(WebsocketClient.prototype, new function() {
     var self = this;
     self.emit('message', JSON.parse(data));
   };
-});
 
+  prototype.getSignature = function(auth) {
```
mihar: Don't need to pass `auth` since it's on the object

lib/clients/websocket.js
```diff
@@ -43,14 +55,20 @@ _.assign(WebsocketClient.prototype, new function() {
   prototype.onOpen = function() {
     var self = this;
     self.emit('open');
+    var auth = self.getSignature(self.auth);
```
mihar: Variable should be called `sig`

lib/clients/websocket.js
CjS77: The 1-liner is actually redundant, so it's removed

PR #21, opened 2016-12-15
Update validation to support different order types

loriopatrick:
Allows size to omitted for market and stop orders.

### DIFF
--- a/lib/clients/authenticated.js
+++ b/lib/clients/authenticated.js
@@ -112,17 +112,35 @@ _.assign(AuthenticatedClient.prototype, new function() {
   prototype._placeOrder = function(params, callback) {
     var self = this;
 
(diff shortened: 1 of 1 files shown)


lib/clients/authenticated.js
```diff
@@ -112,17 +112,35 @@ _.assign(AuthenticatedClient.prototype, new function() {
...
 
-    var requiredParams = ['size', 'side', 'product_id'];
+    if (!params.type) {
+      params.type = 'limit';
+    }
 
-    if (params.type !== 'market') {
-      requiredParams.push('price');
+    if (!params.side) {
+      throw "`opts` must include param `side`";
```
CjS77: Linters complain if you don't throw an `Error`

PR #27, opened 2017-01-16
Enable connecting to multiple products per WS connection

mihar:
Add tests for the WebsocketClient module.

### DIFF
--- a/lib/clients/websocket.js
+++ b/lib/clients/websocket.js
@@ -6,13 +6,13 @@ var signRequest = require('../../lib/request_signer').signRequest;
 
 /**
  * Create a new connection to a websocket feed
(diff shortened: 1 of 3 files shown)


lib/clients/websocket.js
```diff
@@ -90,5 +91,34 @@ _.assign(WebsocketClient.prototype, new function() {
...
+    if (!err) {
+      return;
+    }
+
+    if (err.message === 'unexpected server response (429)') {
+      err = new Error('You are connecting too fast and are being throttled! Make sure you subscribe to multiple books on one connection.');
+      throw err;
+    }
+
+    self.emit(err);
```
CjS77: `self.emit('error', err)` surely

PR #32, opened 2017-03-12
add engines config

givanse:
What Node versions does this npm module support?

I might be wrong, but I'm guessing >= 4. The reason is [Node 0.10, 0.11 and 0.12 have reached end-of-life.](https://github.com/nodejs/LTS#lts-schedule)

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,5 +58,9 @@
   },
   "scripts": {
     "test": "mocha --ui qunit --bail --reporter list tests/*.js"
+  },
+  "engineStrict" : true,
+  "engines": {
+    "node" : ">=4.0.0"
   }
 }


fb55:
The underlying networking library is `request`, as long as they support a node version this library should continue to work as well. (Of course newer versions of node are also supported.)

givanse:
https://github.com/request/request/blob/c57fb7258788888e42fcfa2b6d54c8150517cabe/package.json
```
  "engines": {
    "node": ">= 4"
  },
```

this is great news, most of ES6 could be used ;)

awitherow:
@fb55 any ideas on getting this merged with a higher node verison? Would automagically help a lot of "modernized" failed tests.

nikulis:
Of interest would be Node's [official LTS schedule](https://github.com/nodejs/LTS).

With 4.x reaching end-of-life in less than a year, and 8.x reaching official LTS even sooner than that, I think that going with any of 6.x - 8.x would be an ok move.

In my own opinion, because I love `async`/`await` so much, I would recommend  >= 7.6.0.

fb55:
I merged this to already have a basis for specifying a version. I would like to still support one LTS release, which means we should support node 6. Please let me know if there are any reasons besides async/await* to specify any version higher than 6.0.0.

* I'm also a fan of async/await, but having one battle-tested version of node support this library has precedence for now.

PR #18, opened 2016-12-06
optionally specify productId in authenticated client constructor, passing to public client

mattstabeler:
Edit: as mentioned by @MarcSallent, this solves the issue of not being able to specify a `productId` for methods of `PublicClient` that is inherited by `AuthenticatedClient`

PR #34, opened 2017-03-17
Get product ticker ad hoc through same client

daviscodesbugs:
I needed this feature to be able to interact with each market, without creating a new client connection. Might be of benefit to some users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,10 @@ publicClient.getProductOrderBook({'level': 3}, callback);
 
 * [`getProductTicker`](https://docs.gdax.com/#get-product-ticker)
 ```javascript
+// Get the default product ticker
 publicClient.getProductTicker(callback);
+// Get a specific product ticker by ID
+publicClient.getProductTicker('BTC-USD', callback);
 ```
 
 * [`getProductTrades`](https://docs.gdax.com/#get-trades)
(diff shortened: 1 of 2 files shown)


daviscodesbugs:
Didn't run unit tests beforehand 😳. I'll remake the PR once I know it's working.

awitherow:
@daviscodesbugs apparently you can initiatlize a public Client with a coin/currency. https://github.com/coinbase/gdax-node/pull/18

daviscodesbugs:
@awitherow I recognized this previously, but thought it tedious to maintain multiple clients in order to perform operations such as fetch the most recent ticker.

awitherow:
I have done something like this @daviscodesbugs https://github.com/awitherow/coinbot/pull/19/files#diff-8b7444617c652837169edb6e31a71ddaR7

awitherow:
@daviscodesbugs do you have any ideas where we could have such similar functionality in other coin specific functions?

daviscodesbugs:
@awitherow I don't have a specific use case personally, though it seems that being able to make certain calls, without initializing an additional client, could be useful. Perhaps a client is initialized with a default currency, but subsequent calls could take an optional parameter specifying a different currency.

fb55:
I agree in general that having to initialize a client just to perform a request is quite hazardous. A clients functionality is basically to add a couple of defaults to methods, which could also be passed directly to the method. We might think about making the API surface flat in most cases.

awitherow:
So does this implementation fit your ideal @fb55 ?

fb55:
This is too limited in scope for my taste, with only a single method being updated. I'm closing this PR, but will revisit the topic. @daviscodesbugs I hope that's okay for you!

PR #40, opened 2017-04-03
Modernize codebase

fb55:
`lib/request_signer` was already using a couple of ES6 features, so this shouldn’t break anything.

Switches codebase to classes and arrow functions, replaces lodash with ES5 array extras and normalizes whitespace to four spaces.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,13 +1,7 @@
-var PublicClient        = require('./lib/clients/public.js');
-var WebsocketClient     = require('./lib/clients/websocket.js');
-var AuthenticatedClient = require('./lib/clients/authenticated.js');
-var Orderbook           = require('./lib/orderbook.js');
-var OrderbookSync       = require('./lib/orderbook_sync.js');
-
-module.exports = exports = {
-  'PublicClient'       : PublicClient,
-  'WebsocketClient'    : WebsocketClient,
-  'AuthenticatedClient': AuthenticatedClient,
-  'Orderbook'          : Orderbook,
-  'OrderbookSync'      : OrderbookSync,
+module.exports = {
+    PublicClient: require('./lib/clients/public'),
+    WebsocketClient: require('./lib/clients/websocket'),
+    AuthenticatedClient: require('./lib/clients/authenticated'),
+    Orderbook: require('./lib/orderbook'),
+    OrderbookSync: require('./lib/orderbook_sync'),
 };
--- a/lib/clients/authenticated.js
+++ b/lib/clients/authenticated.js
@@ -1,333 +1,270 @@
-var util = require('util');
-var querystring = require('querystring');
-var async = require('async');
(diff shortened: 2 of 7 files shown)


awitherow:
@fb55  what needs to be done still to get this working?

awitherow:
Currently the test failing is

```
/home/ubuntu/gdax-node/lib/clients/public.js:58
    get(...args) {
        ^^^
```

due to unsupported rest parameters. 

Support for them is shown here: http://node.green/#ES2015-syntax-rest-parameters

They are flagged as warnings below node v6.4.0.

awitherow:
I pulled this branch and set the node engine to `v7.8.0` and all the tests are passing!

The OrderBookSync ones looked a little bit iffy, they took a while to complete.

```javascript
  ✓ OrderbookSync public client should stream trades: 2784ms
  ✓ OrderbookSync public client should stream trades with function: 3365ms
```

awitherow:
ping @fb55

awitherow:
ping @mihar @fb55 ? anyone alive?

awitherow:
@fb55 I would really like to work on getting this merged and working, is there something we can do to get this happening or can the community here get an update at least of what is going on? I have pinged you guys multiple times over the last 2 months... Can we be a little more professional about this?

fb55:
@awitherow I didn't have a chance to investigate the performance regressions at the time, which is why I didn't pursue it any further. I would love it if someone could take this over.

nikulis:
In looking at updating tests for my own refactoring, I found that the current tests for `PublicClient#getProductTradeStream` were actually going to the network / live GDAX API instead of mocking the requests through nock. The variability of the network and live API timing is likely what lead to the difference in execution time.

I've opened a new PR #55 that has a local copy of the requests for nock to use (I used `nock.recorder.rec()` and friends to record the live API calls and save them to a nock config json that could by loaded in tests.)

awitherow:
@fb55 thank you for the response!

Do you guys do any standardized performance testing for gdax-node already, or would we have to create a standard now?

fb55:
@nikulis That's a great find, thanks for looking into it!

@awitherow We don't have anything like that for `gdax-node` right now. The existing tests already help to spot large changes, having some proper benchmarks would be great though.

awitherow:
This would be a learning exercise in terms of how to do benchmark testing & performance stuff... So I am just shooting ideas and needing a little guidance.

Would integration of something like this:

https://benchmarkjs.com

Into the test suite help to track things in the future well?

We could then even write the statistics to json and be able to see the trends in speed of the test suite over time?

Is there a better way to do this?

awitherow:
Also, what is to be done about the node version of the project? With the refactorings, a higher default node version is necessary for the code to even properly function. Thoughts on using v8.0.0?

fb55:
Closing in favor of #60

PR #22, opened 2016-12-27
Initialize eslint

a--hoang:
References #16 .  This is a very simple config file for eslint that tries to follow current conventions on gdax-node.  It produces ~1000 linting errors that I can also include in the PR, if you like.

Btw, hi from one of your future interns!

PR #45, opened 2017-04-25
Remove unused import and fix redeclared error variable

a--hoang:
A little housekeeping - Cleans up the lib/clients/authenticated.js file.


a--hoang:
This is just the code edits (no linting) from #22

PR #47, opened 2017-05-15
Patch 1 - eye candy

bitpup:
corrective toUpperCase method calls removed, authenticated.js & public.js changed accordingly


fb55:
Thanks for this PR! Uppercase names usually signify constants in JS, so while I agree that this could be simplified, this is probably more confusing than helpful.

PR #50, opened 2017-05-25
Added an option to enable the heartbeat option for the websocket.

OFRBG:
I did it for my own project to throttle in a cleaner way. Hope it helps.


awitherow:
Would be nice to see this as a sort of 'option' in a configuration object... to me simply passing true is not very clear.

OFRBG:
Sure. The constructor had two default parameters, one for `auth` and one for the URL. Do I wrap them in the options as well?

awitherow:
Does that make sense to do for you @OFRBG ?

OFRBG:
I think so. Those parameters don't look like usually-modified things, but adding them would wrap it up nicely.

awitherow:
Then go for it @OFRBG ! keep the commit separate for using the config style object so it can be easy to compare then maybe we can get this merged soon.

fb55:
I agree with @awitherow. Boolean arguments are always hard to read, passing an object instead is the right call.

fb55:
Closing in favor of #76

PR #57, opened 2017-06-06
chore(npm): generate package lock

awitherow:
To help better with dependency management.


fb55:
I'm not a fan of having dependency lock files on modules. While they make a lot of sense for products having them for modules will prevent the package manager from de-duplicating dependencies. Also, if we support the npm format we should also support the yarn format, which will become unmanageable at some point.

awitherow:
Good to know!

PR #58, opened 2017-06-06
integrate prettier for *.js code standardization

awitherow:
I saw there was desire in another thread to add prettier, figured I'd go ahead and do so since it is not too difficult 👍 

We need to decide on rules though still.

They can be chosen here: https://github.com/prettier/prettier#options

Merge blocked by PR #40

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,12 +61,12 @@
   },
   "scripts": {
     "test": "mocha --ui qunit --bail --reporter list tests/*.js",
-    "fmt": "node_modules/.bin/prettier --write 'index.js' 'lib/**/*.js' 'tests/**/*.js'",
+    "format": "prettier --single-quote --tab-width 4 --print-width 86 --trailing-comma es5 --write index.js 'lib/**/*.js' 'tests/**/*.js'",
     "precommit": "lint-staged"
   },
   "lint-staged": {
     "*.js": [
-      "node_modules/.bin/prettier --write",
+      "prettier --write",
       "git add"
     ]
   }
--- a/package.json
+++ b/package.json
@@ -23,6 +23,8 @@
   "dependencies": {
     "async": "1.5.0",
     "bintrees": "1.0.0",
+    "husky": "^0.13.4",
+    "lint-staged": "^3.6.0",
     "lodash.assign": "3.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -57,6 +60,14 @@
     "url": "git://github.com/coinbase/gdax-node.git"
   },
   "scripts": {
-    "test": "mocha --ui qunit --bail --reporter list tests/*.js"
+    "test": "mocha --ui qunit --bail --reporter list tests/*.js",
+    "fmt": "node_modules/.bin/prettier --write 'index.js' 'lib/**/*.js' 'tests/**/*.js'",
```
fb55: Let's call it `format`. Also, `node_modules/.bin/` is unnecessary.

fb55: Also, we're currently using four spaces in most places, so let's use that. Also, we want es5 dangling commas.

package.json
```diff
@@ -57,6 +60,14 @@
...
+    "precommit": "lint-staged"
+  },
+  "lint-staged": {
+    "*.js": [
+      "node_modules/.bin/prettier --write",
```
fb55: ~~`node_modules/.bin/`~~

package.json
fb55: There also isn't a need for quotes around `'index.js'`, and we should be okay just looking for all js files.

fb55:
Thanks a lot for this PR! Not sure whether we shouldn't wait for #40 with this, as it will introduce a lot of merge conflicts.

awitherow:
@fb55 yeah sure thing! I can remove the refactor commit and then we can try to get #40 cleaned up and then we can run it again after merging.

package.json
awitherow: I was not 100% that one could use prettier if not installed globally.

package.json
awitherow: 👍 thanks for the comments, hadn't officially set up prettier in a project before, only was using the vscode plugin on pre-existing setups.

fb55:
Fixed with #60. Thanks a lot @awitherow for getting this started!

awitherow:
🙇 my pleasure!

PR #55, opened 2017-06-06
Add nock requests for publicClient.getProductTradeStream() tests

nikulis:

Issue #41, opened 2017-04-06
Add support for promises


nikulis:
Are there currently any plans to add promise support in the API clients? It would be very nice for those who like to write promise-style rather than use callbacks.

To use an example from the README, the following code:

```js
var callback = function(err, response, data) {
  // your code here.
};

authedClient.getOrders({'after': 3000}, callback);
```

…could instead be used as:

```js
authedClient.getOrders({'after': 3000})
  .then(result => {
    let { data, response } = result; // Still expose underlying response object?
    /* do stuff with data */
  })
  .catch(err => { /* handle error */ });
```

Furthermore, returning a promise would also allow for use with async/await, e.g.

```js
async function myFunction() {
  try {
    let { data, response } = await authedClient.getOrders({'after': 3000});
    /* do stuff with data */
  } catch (err) { /* handle error */ }
}
```

From an implementation perspective, I think it would be totally possible for the same functions to support callbacks and promises at the same time, without making any breaking changes. For example, existing functions could run callbacks as normal if provided and return a promise if no callback is provided. Or, as the current API call functions don't return anything, they could just always return a promise (but still execute a callback if provided).

I haven't been able to find any documentation about minimum versions of Node that this project must/wants to support, but with PR #40 starting to overhaul and ES6-ify things, I'm guessing that using native ES6 promises would be fine platform-support-wise.

I also haven't seen any contributing guidelines, but I'd be happy to pitch in and help implement this if it's something more people would like.

awitherow:
This would be super ideal!

PR #60, opened 2017-06-09
Promisify

nikulis:
Here's my first pass at adding in promise support re: #41.

The main area of interest for this PR is the `tests` directory (re: discussion on #55), where I've reorganized the test suites and added tests that exemplify uses of both callbacks and promises where applicable.

When I started work on this, I had not yet realized that the blocking performance regressions on #40 were due to how tests were running instead of introduced code complexity, so I did my own port to ES2015/17 (also just to get myself familiar with the codebase). I don't mean to replace the hard work others have done on #40, and would be happy to port my changes over to that branch after the maintainers have decided on Node version compatibility and updated the CI to build accordingly.


lib/clients/authenticated.js
```diff
@@ -1,332 +1,240 @@
...
-    prototype[method] = _.partial(prototype.request, method);
-  });
+  }
 
-  prototype.getAccounts = function(callback) {
-    var self = this;
-    return prototype.get.call(self, ['accounts'], callback);
-  };
+  getAccounts(callback) {
+    return this.get(['accounts'], callback);
```
awitherow: thoughts on using arrow syntax in classes?

lib/clients/authenticated.js
```diff
@@ -1,332 +1,240 @@
...
 
-  prototype.cancelOrder = function(orderID, callback) {
-    var self = this;
+  cancelOrder(orderID, callback) {
 
-    if (!callback && (typeof orderID === 'function')) {
-      callback = orderID;
-      callback(new Error('must provide an orderID or consider cancelOrders'));
-      return;
+    if(!orderID || typeof orderID == 'function') {
```
awitherow: Is the typeof 'function' check necessary here?

lib/clients/authenticated.js
awitherow: thoughts on using arrow syntax in classes?

lib/clients/authenticated.js
```diff
@@ -1,332 +1,240 @@
...
-    var currentDeletedOrders = [];
-    var totalDeletedOrders = [];
-    var query = true;
-    var response;
+    return this.get(['products', productId, 'book'], { qs: args }, callback);
+  }
 
-    args = args || {}
-    if (!callback && (typeof args === 'function')) {
+  async cancelAllOrders(args = {}, callback) {
```
awitherow: 😍

lib/clients/authenticated.js
```diff
@@ -1,332 +1,240 @@
...
+        });
 
-        if ((resp && resp.statusCode != 200) || !data) {
-          var error = new Error('Failed to cancel all orders');
-          query = false;
-          done(error);
-          return;
-        }
+        totalDeletedOrders.push(...data);
+      } while (data.length);
```
awitherow: elegant while!

lib/clients/authenticated.js
```diff
@@ -1,332 +1,240 @@
...
 
-  prototype.getOrder = function(orderID, callback) {
-    var self = this;
-
-    if (!callback && (typeof orderID === 'function')) {
-      callback = orderID;
-      callback(new Error('must provide an orderID or consider getOrders'));
-      return;
+  getOrder(orderID, callback) {
+    if (!orderID || typeof orderID == 'function') {
```
awitherow: again on the use of the orderID function check.

lib/clients/public.js
```diff
@@ -1,206 +1,213 @@
...
-  prototype.getProductOrderBook = function(args, callback) {
-    var self = this;
+  getProducts(callback) {
+    return this.get(['products'], callback);
+  }
 
-    args = args || {}
-    if (!callback && (typeof args === 'function')) {
+  getProductOrderBook(args = {}, callback) {
+    if (!callback && typeof args === 'function') {
```
awitherow: These couple lines of code are duplicated so much in this repo in general, thoughts on what could be done to reduce that in a functional way?

lib/clients/public.js
```diff
@@ -1,206 +1,213 @@
...
+        }
 
-      for (var i = data.length - 1; i >= 0; i--) {
-        if (shouldStop && shouldStop(data[i])) {
-          stream.push(null);
+        if (resp.statusCode === 429) {
+          // rate-limited, try again
+          setTimeout(() => {
+            fetchTrades.call(this, stream, tradesFrom, tradesTo, shouldStop);
+          }, 900);
```
awitherow: why 900? Just curious about the setTimeout.

lib/clients/websocket.js
```diff
@@ -1,111 +1,87 @@
...
+  disconnect() {
+    if (!this.socket) throw new Error('Could not disconnect (not connected)');
+    this.socket.close();
+  }
 
-  prototype.onOpen = function() {
-    var self = this;
-    self.emit('open');
+  onOpen() {
+    this.emit('open');
```
awitherow: confused at this here... `onOpen` is called when 'open' is emitted, and then it immediately emits 'open' again?

lib/clients/authenticated.js
nikulis: How so? If you mean something like

```js
class Whatever {
  getAccounts: callback => this.get(['accounts'], callback);
}
```

I'm pretty sure that's invalid JS syntax.

lib/clients/authenticated.js
nikulis: I think their original intent was to ensure that the first argument was an orderID instead of a callback. So if `orderID` were a function, then `!orderID === true`, so you have to also check to make sure it's not a function.

lib/clients/public.js
nikulis: Not sure, that's the timeout that was used previously.

lib/clients/public.js
nikulis: In terms of optimizing syntactic aesthetics, decorators are the first thing I think of,

```js
@optionalArgs
someFunction(args, otherParam, callback) {
  /* ... */
}
```

But even though [decorators are in Stage 2](https://tc39.github.io/proposal-decorators/), there isn't great out-of-the-box support for transpiling them quite yet, including in the latest version of [babel](https://babeljs.io/docs/plugins/transform-decorators/).

lib/clients/websocket.js
nikulis: Hmm, now that you point that out, I'm confused as well. This was the previous behavior. Thoughts @fb55?

lib/clients/authenticated.js
awitherow: ```func = prop => this.otherfunc()``` works?

lib/clients/authenticated.js
awitherow: ahh yeah right, I didn't think about it enough 👍

lib/clients/public.js
awitherow: Yeah definitely something for the future 🌈 🦄

lib/clients/public.js
awitherow: Hmm interesting. 👍

lib/clients/authenticated.js
nikulis: Not until [ES Class Fields](https://github.com/tc39/proposal-class-fields) are finalized (currently stage 2)

awitherow: Jumping between babelified React Native, React and Node can get confusing 😅

nikulis: I feel your pain!

.eslintrc.json
```diff
@@ -0,0 +1,21 @@
+{
```
fb55: This should probably extend eslint's recommended settings (`eslint:recommended`).

lib/clients/authenticated.js
```diff
@@ -1,332 +1,240 @@
...
 
-        if (err) {
-          done(err);
-          return;
-        }
+    try {
+      do {
+        // Have to use a special promise to get response as well as data
+        // for later passing to the callback;
+        [response, data] = await new Promise((resolve, reject) => {
```
fb55: I'd prefer to not use async/await in this library for now (node@6 should be the target).

lib/clients/authenticated.js
```diff
@@ -1,332 +1,240 @@
...
+          this.delete(['orders'], opts, (err, response, data) => {
+            if (err) reject(err);
+            resolve([response, data]);
+          }).catch(() => {}); // silently ignore, rejection is already handled in callback
```
fb55: Ideally this would be appended every time we pass a callback.

lib/clients/authenticated.js
```diff
@@ -1,332 +1,240 @@
...
+  request(method, uriParts, opts = {}, callback) {
+    if (!callback && typeof opts == 'function') {
       callback = opts;
       opts = {};
     }
 
-    if (!callback) {
-      throw "Must supply a callback"
-    }
+    if (!callback) callback = () => {};
```
fb55: Let's handle this in `makeRequestCallback`

lib/clients/public.js
```diff
@@ -1,206 +1,213 @@
...
-    var started = false;
+    const rs = new Readable({ objectMode: true });
+    let started = false;
 
-    rs._read = function() {
+    rs._read = () => {
       if (!started) {
         started = true;
-        fetchTrades(self, rs, tradesFrom, tradesTo, shouldStop, 0);
+        fetchTrades.bind(this)(rs, tradesFrom, tradesTo, shouldStop, 0);
```
fb55: `.call`

lib/clients/public.js
fb55: The rate limit is per second, this skips to the next second. Not sure if this is the ideal solution, we could also just trigger a couple more requests.

lib/clients/websocket.js
```diff
@@ -1,111 +1,87 @@
-var EventEmitter = require('events').EventEmitter;
-var Websocket = require('ws');
-var Utils = require('../utilities.js');
-var util = require('util');
-var _  = {assign: require('lodash.assign')};
-var signRequest = require('../../lib/request_signer').signRequest;
+const EventEmitter = require('events').EventEmitter;
+const Websocket = require('ws');
+const Utils = require('../utilities.js');
+const signRequest = require('../../lib/request_signer').signRequest;
```
fb55: Destructuring for `EventEmitter` and `signRequest` — should also be enforced by eslint (`prefer-destructuring`).

nikulis:
Hmm, CircleCI seems to have independently fallen down on this one (["An error occurred while fetching information about this build from GitHub."](https://circleci.com/gh/coinbase/gdax-node/92)).

I spun up my own Circle instance and it seems to be [working fine](https://circleci.com/gh/nikulis/gdax-node/2), though. Can anyone with access go in and trigger a re-build?

nikulis:
Side note: It seems that Circle [doesn't read node engine settings from `package.json`](https://discuss.circleci.com/t/use-node-version-from-package-json/122), so we have to use a `circle.yml` config file for now.

nikulis:
Just checking in - would anyone have time to review these newer changes?

If all is well, I propose merging this branch and getting a new release out very soon, possibly with some other addressed issues as well. Thinking towards the future, I also still have nikulis/gdax-node#1 open (based on this branch) which implements the more standard two-argument callback API (major breaking change).

There are many other open issues that I would like to work on, but I am trying to limit the scope of this PR itself. If the maintainers are interested in moving forward with this PR (as is my impression), I think it would be valuable to get it through soon; there seem to be many community members submitting PRs based off of the current master which would all need significant refactoring in order to apply here, and I think it would be fairer to them to not have to duplicate so much work going forward. That being said, I'm always happy to help with related tasks too :)

Thanks!

tests/websocket.spec.js
```diff
@@ -1,38 +1,45 @@
...
         server.close();
         done();
       });
     });
   });
 
-  test('subscribes to the default product (BTC-USD) if undefined', function(done) {
-    var server = testserver(++port, function() {
-      var websocketClient = new Gdax.WebsocketClient(null, 'ws://localhost:' + port);
+  test('subscribes to the default product (BTC-USD) if undefined', function(
```
fb55: All of these should be switched to arrow functions (eslint can also take care of that — only named functions should still be allowed).

.eslintrc.json
```diff
@@ -0,0 +1,24 @@
...
+  "root": true,
+  "extends": "eslint:recommended",  
+  "env": {
+    "commonjs": true,
```
fb55: Pretty sure this is implied by `node`

.eslintrc.json
```diff
@@ -0,0 +1,24 @@
...
+    "es6": true,
+    "node": true
+  },
+  "parserOptions": {
+    "sourceType": "module",
+    "ecmaVersion": 2015
+  },
+  "rules": {
+    "constructor-super": "warn",
```
fb55: All of these should error.

README.md
```diff
@@ -12,186 +12,277 @@ the MIT License – please take the following message to heart:*
...
+    // work with data
+  })
+  .catch(error => {
+    // handle the error
+  });
 ```
-var callback = function(err, response, data) {
-  // your code here.
-};
+
```
fb55: Perhaps add some description here: "Promises can be used with callbacks, as well as using `async/await` in modern environments." (or something similar).

lib/clients/public.js
```diff
@@ -1,206 +1,214 @@
...
       }
-      callback(err, response, data);
+      if (typeof callback == 'function') {
+        callback(err, response, data);
+        return;
+      }
+      if (err) reject(err);
+      if (data === null)
+        reject(new Error('Response could not be parsed as JSON'));
+      else resolve(data);
```
fb55: I've been looking into replacing parts of this library with [swagger-codegen](https://github.com/swagger-api/swagger-codegen). Their [promise interface](https://github.com/swagger-api/swagger-codegen/blob/0416b0332f36b4a7f80a4313df8f8257a4539fe4/modules/swagger-codegen/src/main/resources/Javascript/es6/ApiClient.mustache#L462) returns a `{data, response}` object — I guess we should adopt that as well. While I like the simplicity of this API, there probably are cases where you want the request object (for headers, status codes etc.), plus an object would allow for future extensions.

lib/clients/public.js
```diff
@@ -1,206 +1,225 @@
...
+    const p = new Promise((resolve, reject) => {
+      request(opts, this.makeRequestCallback(callback, resolve, reject));
+    });
 
-  prototype.getProducts = function(callback) {
-    var self = this;
-    return prototype.get.call(self, ['products'], callback);
-  };
+    if (callback) {
+      p.catch(() => {});
```
fb55: This is definitely sensible behaviour, although I'd simply add listeners to the promise in case there is a callback. Still return the original promise (not the one with the callbacks applied) so that nothing needs to be returned from the callback.

We could also get rid of `makeRequestCallback` in favour of superagent promises — how do you feel about that?

fb55:
One comment, but that can be iterated upon once this is merged. There is a merge conflict with the `package.json` file — once that's resolved I'm happy to merge. Thanks a lot for all the work, this is amazing.

fb55:
Finally got around to merging this. Thanks a lot @awitherow and especially @nikulis! I'd like to land the superagent switch before publishing it — @nikulis do you want to open a PR with the code you already have?

awitherow:
Thanks to @nikulis I was just giving some insight! He did the 💪. 👏

nikulis:
I was holding off on superagent for this one because those using v0.4.x might be relying upon the currently-exposed `response` object (2nd callback parameter) that the `request` library returns. I've personally never used this in my own projects, so I'd have to look into whether the `response` object could be easily reconstructed using the information that superagent returns. If so, I can 

I *did* switch to superagent with the new two-parameter callback API in nikulis/gdax-node#1. That branch is a little stale atm, so I'm about to work on getting it up-to-date before I open a new PR in this repo. (Just got back online after a few days off 😅 )

And yes, thank you to all of you, too, for providing all the wonderful feedback and critiques!

PR #68, opened 2017-06-19
Refactored tests/added new tests

### LINKED ISSUE #63: getProductTrades swallows errors
You should implement the [callback pattern](https://blog.yld.io/2015/10/19/the-callback-pattern/#.WT3MVRPyuAx), it makes a developers life a lot easier and i would argue it is good practice to do in a crazy untyped language like javascript.

a callback should be of the form:
`function callback(error,result) {
}`

If my understanding is correct:
The first callback argument is null or undefined;
The second argument contains the callback results.
On the other hand, if an error occurred:
the first argument contains an error object, an instance of the JavaScript Error class (or sub-class) that describes the error;
the second argument will be undefined.

Your getProductTrades function at least, is returning null for the error parameter regardless of a success of failure. So what is even it's use and why is it mentioned as part of your specific callback?

Miladinho:
Hi,

I am trying to fix #63 this bug and I noticed that the tests in tests/public_client.js are very insufficient to begin with and actually do not cover many of the uses cases correctly. To begin with, I added tests that would make sure the error object in the callback to PublicClient.getProductTrades behaves as expected, because as noted above #63 it does not do so currently. 

tests/orderbook_sync_test.js
I noticed that my tests were not behaving correctly because an Error object was being returned in the first test, but a quick console.log(err) in the test showed that my test's getProductTrades is still calling the nock mock which was weird because I did not use nock. 
`․ OrderbookSync builds specified books: 13ms
    PublicClient PublicClient with bad market input should return error on call to getProductTrades: { [Error: Nock: No match for request GET https://api.gdax.com/products/non-existant%20market/trades ] status: 404, statusCode: 404 }`

I figured that 'OrderbokSync builds specified books' suite was not clearing nock and I was correct so I added the line 'nock.clearAll();' (other files have tests that miss this line as well but I will take care of that in other commits)

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,5 +58,9 @@
   },
   "scripts": {
     "test": "mocha --ui qunit --bail --reporter list tests/*.js"
+  },
+  "engineStrict" : true,
+  "engines": {
+    "node" : ">=4.0.0"
   }
 }
--- a/tests/orderbook_sync_test.js
+++ b/tests/orderbook_sync_test.js
@@ -65,6 +65,7 @@ describe('OrderbookSync', function() {
 
     server.on('connection', function(socket) {
       server.close();
+      nock.clearAll();
       done();
     });
   });


Miladinho:
The check that failed is 'npm test' and that is supposed to fail, because the tests I added are highlighting faulty functionality in current production. Once approved and logged I will add the changes that will pass the test. I am using the TDD process.

fb55:
@Miladinho Sorry for leaving this hanging — could you please verify whether this was already fixed, and rebase this PR?

Issue #24, opened 2016-12-30
OrderbookSync does not reconnect when socket connection is lost


petercerno:
When the socket connection is lost, `OrderbookSync` ends up in a stale state (it does not attempt to re-connect and thus also stops updating the underlying orderbook).

I have "solved" this issue in my app by re-connecting the orderbook whenever the last update is older than few minutes. However, it would be better if this was fixed in the master branch.

ctubio:
```ts
  constructor(this._client) {
    this._client.on("close", data => this.onStateChange());
  }

  private onStateChange = () => {
    if (!this._client.socket) setTimeout(() => this._client.connect(), 5000);
  };
```

fb55:
@ctubio Mind opening a PR?

julesGoullee:
why this issue([bug]) is open ?

rmm5t:
Closing this due to inactivity. Please open a new Issue and/or PR (possibly referencing this issue) if further action is deemed necessary.

PR #72, opened 2017-06-23
Fixed Cannot read property 'sub' of undefined.

ctubio:
Fix for bug introduced at https://github.com/coinbase/gdax-node/commit/d85184e2c844fdbba5fc7a5182fce9294bd6197c

this just recovers old logic before new logic throws an error (Cannot read property 'sub' of undefined).

:gift_heart:

### DIFF
--- a/lib/orderbook.js
+++ b/lib/orderbook.js
@@ -137,7 +137,10 @@ _.assign(Orderbook.prototype, new function() {
       return order.id === match.maker_order_id;
     });
 
-    assert(order);
+    if (order === -1 ) {
+      order = node.orders[0];
+      assert.equal(order.id, match.maker_order_id);
+    } else assert(order);
 
     order.size = order.size.sub(size);
     self._ordersByID[order.id] = order;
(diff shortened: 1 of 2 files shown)


ctubio:
well, fixes comes in pairs, the second is because https://github.com/coinbase/gdax-node/issues/24

lib/orderbook.js
```diff
@@ -137,7 +137,10 @@ _.assign(Orderbook.prototype, new function() {
       return order.id === match.maker_order_id;
     });
 
-    assert(order);
+    if (order === -1 ) {
```
awitherow: Can you explain the -1? I am not sure how this is working.

ctubio: `find` must return null on record not found (see https://github.com/vadimg/js_bintrees/blob/473c12feeb91ecb5aa4c8eaf35c4143fa84ede46/dist/rbtree.js#L25), i dont know why it returns -1, but it does; and -1 doesn't have any property `size` with a `sub` method.

ctubio: lets close this meanwhile someone knows why returns -1

PR #73, opened 2017-06-23
Fixed Cannot read property 'sub' of undefined.

ctubio:
refix of https://github.com/coinbase/gdax-node/pull/72

donno why for me and my users `find` returns -1 even if `node.orders` have an order with `id` equal to `match.maker_order_id`; this PR just does a `find`using `forEach`, that does not return -1 never

:gift_heart:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "gdax",
-  "version": "0.3.1",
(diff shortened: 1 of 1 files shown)


ctubio:
i feel a much better fix to simply recover 4.2

PR #76, opened 2017-07-07
Heartbeat

amejiarosario:
Adds the [heartbeat](https://docs.gdax.com/#heartbeat) option to Websocket


amejiarosario:
Could any of you please review this changes?
@awitherow @gcperrin @givanse Thanks!

lib/clients/websocket.js
```diff
@@ -8,17 +8,21 @@ var signRequest = require('../../lib/request_signer').signRequest;
...
- * @param auth {object} Optional. An object containing your API ket details (key, secret & passphrase)
+ * @param options {object} Optional. It can contain auth, websocketURI and heartbeat
+ *  - auth {object} Optional. An object containing your API ket details (key, secret & passphrase)
+ *  - websocketURI {string} Optional. The websocker URL. Default: The official GDAX feed.
+ *  - heartbeat {boolean} Optional. Receive heartbeat messages once a second
  */
-var WebsocketClient = function(productIDs, websocketURI, auth) {
+var WebsocketClient = function(productIDs, options) {
   var self = this;
+  options = options || {};
```
awitherow: Due to the breaking changes in the way the `WebsocketClient` takes parameters, do you think it would be good to add the following more parameter error handling?

For example, maybe checking to ensure that `options` is in fact an object, and not a string (perhaps websocketURI).

README.md
```diff
@@ -316,6 +316,12 @@ var Gdax = require('gdax');
 var websocket = new Gdax.WebsocketClient(['BTC-USD', 'ETH-USD']);
 websocket.on('message', function(data) { console.log(data); });
 ```
+
+Optionally set the heartbeat mode.
+```javascript
+var websocket = new Gdax.WebsocketClient(['BTC-USD','ETH-USD'], {heartbeat: true});
```
awitherow: Here it would be good to expand upon the optional parameters (websocket URI, authentication) just so that people know more without having to dig into the code themselves to figure out how it works, or why their old implementation is breaking if they've upgraded.

lib/clients/websocket.js
awitherow: This is not the only case either, just a thought of one potential.

The idea is to look at this and ensure it is as backwards compatible friendly as possible, without making it mess.

fb55:
@amejiarosario This needs a rebase, great patch otherwise.

README.md
amejiarosario: ok, I'll update

amejiarosario:
@awitherow made it backward compatible.
@fb55 rebased!

fb55:
@amejiarosario Thanks a lot! We should update this to use the `heartbeat` channel, but I didn't want to block this any further.

Issue #11, opened 2016-10-05
"Size" parameter always required for Market orders


simlay:
According to https://docs.gdax.com/#place-a-new-order:

```
MARKET ORDER PARAMETERS

Param   Description
size    [optional]* Desired amount in BTC
funds   [optional]* Desired amount of quote currency to use

* One of size or funds is _required.
```

So, the following should work to place an order:

```
var buyParams = {
    'funds': 1, // USD,
    'currency': 'BTC',
    'type': 'market',
};
authedClient.buy(buyParams, callback);
```

Anyway, it looks like [size is always required](https://github.com/coinbase/gdax-node/blob/57afad7f7c401ccc6da36d273f14703f0c8a8ea0/lib/clients/authenticated.js#L121-L131) rather than just one of `size` or `funds`.

poffdeluxe:
I put up a simple PR for this guy.

PR #77, opened 2017-07-08
Allow for just funds to be passed in for market orders

poffdeluxe:
Fixes https://github.com/coinbase/gdax-node/issues/11

### DIFF
--- a/lib/clients/authenticated.js
+++ b/lib/clients/authenticated.js
@@ -111,7 +111,7 @@ _.assign(AuthenticatedClient.prototype, new function() {
   prototype._placeOrder = function(params, callback) {
     var self = this;
 
-    var requiredParams = ['size', 'side', 'product_id'];
+    var requiredParams = ['side', 'product_id'];
 
     if (params.type !== 'market') {
       requiredParams.push('price');
--- a/lib/clients/authenticated.js
+++ b/lib/clients/authenticated.js
@@ -115,6 +115,7 @@ _.assign(AuthenticatedClient.prototype, new function() {
 
     if (params.type !== 'market') {
       requiredParams.push('price');
+      requiredParams.push('size');
     }
 
     _.forEach(requiredParams, function(param) {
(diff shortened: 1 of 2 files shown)


awitherow:
Is there anything in the documentation (readme.md) about these changes you've made?

lib/clients/authenticated.js
```diff
@@ -111,17 +111,24 @@ _.assign(AuthenticatedClient.prototype, new function() {
   prototype._placeOrder = function(params, callback) {
     var self = this;
 
-    var requiredParams = ['size', 'side', 'product_id'];
+    var requiredParams = ['side', 'product_id'];
 
     if (params.type !== 'market') {
       requiredParams.push('price');
+      requiredParams.push('size');
```
awitherow: What happens when the params type is market and size is passed into the params anyways?

poffdeluxe: In addition to a `funds` parameter? My understanding according to the API reference is that size will be used to place the order but it'll limit the buy according to funds.

>When placing a market order you can specify funds and/or size. Funds will limit how much of your quote currency account balance is used and size will limit the bitcoin amount transacted.

poffdeluxe:
Yep! I'll try to expand documentation and tests a lil bit. Will put something together this week. May need to go back and forth with y'all to figure out the best way to format/expand the docs.

poffdeluxe:
Also gonna squash my branch and fix my commit messages

lib/clients/authenticated.js
awitherow: 👍 I was just going through and doing some code reviews for notifications. I am not super knowledgable about the functionality there so I trust you 😃

fb55:
Fixed in 93f19ac

fb55:
Thanks a lot @poffdeluxe for the PR!

fb55:
Also, feel free to open another PR if you get around to extend docs and/or tests

poffdeluxe:
@fb55 Will do!

PR #78, opened 2017-07-16
Fix JSDoc & typos

stoivan:


fb55:
Thanks a lot @stoivan!

PR #79, opened 2017-07-19
Require `productID` to be passed to method.

passabilities:
Don't require the productID to be set on creating the client. This is to prevent creating multiple unnecessary client instances if working with multiple pairs.


fb55:
This is a breaking change without anything gained from it. It might make sense to allow specifying the product ID on method calls, but clients encapsulating product & authentication information definitely has its value.

Issue #80, opened 2017-07-23
OrderbookSync returns stale data


petercerno:
Hi, when creating `OrderbookSync` as follows:

```
this._orderbookSync = new Gdax.OrderbookSync(
      productIds, apiURI, websocketURI, this._authClient);
```

And then reading the `bid` / `ask` prices as follows:

```
let state = this._orderbookSync.books[productId].state();
let bid = state.bids[0].price;
let ask = state.asks[0].price;
```

What happens is that `bid` or `ask` prices stop being updated (after some time) and become stale (inconsistent with what I see on https://www.gdax.com/trade/). This happens randomly, but often enough, that I have to re-create `OrderbookSync` almost every minute. Is there a bug? (It used to work fine before).

I have used `ETH-EUR` product.

pranksterN1:
PR #81 did't fix this issue  :(

pranksterN1:
There is warning [here](https://docs.gdax.com/#real-time-order-book):
> All open and match messages will always result in a change to the order book. Not all **done** or **change** messages will result in changing the order book. These messages will be sent for received orders which are not yet on the order book. **Do not alter the order book for such messages**, otherwise your order book will be incorrect.

But [here](https://github.com/coinbase/gdax-node/blob/8611f3628eef1d6d2e9e83d1b7218a359854c428/lib/orderbook_sync.js#L113) I can see that "done" and "change" messages are processed in orderbook. Is this correct ?

gmelika:
I think this comment might explain the stale data issue:

https://github.com/coinbase/gdax-node/issues/104#issuecomment-327882166

fb55:
This library refetches orders when it reconnects. Not sure what the exact origin of this is, especially as this seems limited on one side of the book. I hope to update gdax-node to use our level 2 channel soon, which should fix this issue. (See #82.)

ztnark:
I am seeing this issue in 0.6.0. Is there a solution, or is OrderBookSync unusable? Thank you.

johanhenriksson:
Im seeing this problem in 0.8.0 as well. Asks stop updating after a while, bids keep updating as normal. Are there any known workarounds?

BGrades:
I am seeing the same problem in 0.8.0. I have the sync orderbook created with 4 productIds. And as @johanhenriksson described the asks stops updating after some time. Worth pointing out that the asks are stoping to update one by one. Does anyone know the reason for that or have a workaround? Thank you.

[...]

Issue #62, opened 2017-06-11
Version 0.4.3 orderbook isn't kept updated


kenfehling:
I've created two test repos, one with 0.4.2 and one with 0.4.3 (master). They each print out the top bid and ask every few seconds.

[0.4.2 repo](https://github.com/kenfehling/gdax-orderbook-test-0.4.2)
[0.4.3 repo](https://github.com/kenfehling/gdax-orderbook-test-0.4.3)

I recorded results for each started at approximately the same time, not exactly so the values are slightly different, but the main finding is that after about a minute the 0.4.3 version begins to get stale on one side of the order book (in this instance the ask side, not sure if that's always the case).

0.4.2
```
2963.87 2962.96
2963.91 2962.94
2963.94 2963.86
2963.94 2963.28
2963.86 2959.02
2964.97 2961.73
2964.98 2964.21
2964.96 2964.2
2964.94 2964.16
2965.6 2964.61
```

0.4.3
```
2963.9 2962.51
2963.88 2962.51
2963.9 2963.79
2963.9 2963.79
2963.88 2962.51
2963.9 2962.51
2963.9 2964.21
2963.9 2964.2
2963.9 2964.16
2963.9 2964.55
```

kenfehling:
Now I'm running my tests again and it's fine. Sorry if this is an issue on my end. I'll keep researching.

kenfehling:
Actually it seems like the issue might happen when I use the new multiple product feature. I've updated the repo. Also it looks like it's not completely stale, just a lot slower to update, and it happens on both sides of the order book.

awitherow:
Can you please let us know what node version and npm version you are using?

@fb55 can we do some sort of minimum issue template where creators have to input such rudimentary information?

kenfehling:
node: v6.10.2
npm: 3.10.10

awitherow:
Thanks @kenfehling the node/npm version is maybe for future users or for further debugging info.

Rampage89:
I have experienced the same issue as well, one side of the orderbook is not kept up to date, mostly the asks..

ambethia:
Do we think this is an issue with the `gdax-node` client itself or are the messages coming over the websockets connection delayed?

kenfehling:
My guess is whatever the issue is, it started at some point after commit 98d044efba8329e0fd4f583ae2efaaf60ac81d8c on April 3 because it appears to happen on 0.4.3 but not 0.4.2

[...]

PR #81, opened 2017-07-23
Delayed initial orderbook sync until first message is received (#62).

rsparlin:
The orderbook sync issue happens when the initial loadOrderbook finishes before the websocket client has subscribed to any products, which will cause some initial updates to be missed. This can also cause several unnecessary resyncs. Waiting until after the first message for that product is received resolved the issue (#62, #80) in my testing. A setTimeout for several seconds on the initial loadOrderbook call works most of the time as well, but it is not guaranteed.

### DIFF
--- a/lib/orderbook_sync.js
+++ b/lib/orderbook_sync.js
@@ -21,9 +21,8 @@ class OrderbookSync extends WebsocketClient {
 
     this.productIDs.forEach(productID => {
       this._queues[productID] = [];
-      this._sequences[productID] = -1;
+      this._sequences[productID] = -2;
       this.books[productID] = new Orderbook();
-      this.loadOrderbook(productID);
     });
   }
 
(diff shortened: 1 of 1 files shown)


fb55:
@rsparlin Good idea, would you mind rebasing this?

Also, just curious: What's your opinion on #82?

rsparlin:
Done. However, I did notice an unrelated issue that made orderbook syncing fail unless I corrected it. Line 79 of orderbook_sync.js needed `this.processMessage` to have `this` bound to it. I can separate this into a different PR if you like, but I have included the change for now.

As for #82, I would welcome the change; I can't think of a situation where I've been concerned with specific orders other than my own.

lib/orderbook_sync.js
```diff
@@ -71,7 +76,7 @@ class OrderbookSync extends WebsocketClient {
       this.books[productID].state(data);
 
       this._sequences[productID] = data.sequence;
-      this._queues[productID].forEach(this.processMessage);
+      this._queues[productID].forEach(this.processMessage.bind(this));
```
fb55: Good catch. The context can also be passed as an additional argument to `forEach`, so we'd end up with `this._queues[productID].forEach(this.processMessage, this);`.

fb55:
Decided to fix the nit in #83. Thanks a lot @rsparlin for this, should make the orderbook sync much more reliable.

Issue #46, opened 2017-04-29
Authenticated getProductOrderBook overridden


pgray64:
Is there a particular reason why the authenticated getProductOrderBook is overridden as function(args, productId, callback)? The change of arguments from the public version getProductOrderBook = function(args, callback) is not documented anywhere but the source.

nikulis:
I agree that the authenticated client should not override this.

A [blame](https://github.com/coinbase/gdax-node/blame/cdea38a18b0609aed97b877483444d5ea975cd32/lib/clients/authenticated.js#L163) reveals that the code was inserted with a comment describing it as a "temp override" in faece04 as part of coinbase/coinbase-exchange-node#35 (no description provided).

It was then used in [`bf6c2a4`](https://github.com/coinbase/gdax-node/commit/bf6c2a4503032ce76482aba389817040323c1fd2#diff-b8485adf65396e9a8815ffa587a1ce96R51) to add support for `OrderbookSync` to use an existing authenticated client as part of coinbase/coinbase-exchange-node#37 (no description provided).

I'm not sure as to the rationale of why these changes were introduced.

More recently, #37, which added support for multiple order books (specifically [2fd3464](https://github.com/coinbase/gdax-node/pull/37/commits/2fd3464a21da9a575f1a2656d36168cd9b04c4e0#diff-b8485adf65396e9a8815ffa587a1ce96R64)), seems to have taken a new approach by memoizing `PublicClient`s for each requested book type, which seems like a better way to go imho.

Tagging relevant committers who may have some more insight:
@cilphex, @maxim-s, @fb55

PR #83, opened 2017-07-26
Add productId to authenticated clients & some cleanup

fb55:
fixes #18

### DIFF
--- a/lib/clients/authenticated.js
+++ b/lib/clients/authenticated.js
@@ -7,7 +7,7 @@ class AuthenticatedClient extends PublicClient {
     key,
     b64secret,
     passphrase,
-    apiURI = 'https://api.gdax.com',
+    apiURI,
     productID = null
   ) {
     super(productID, apiURI);
@@ -134,24 +134,6 @@ class AuthenticatedClient extends PublicClient {
     return this.delete(['orders'], callback);
   }
 
-  // temp over ride public call to get Product Orderbook
-  getProductOrderBook(args = {}, productId, callback) {
-    if (!callback) {
-      if (typeof args !== 'object') {
-        callback = productId;
-        productId = args;
-        args = {};
-      }
-
-      if (typeof productId === 'function') {
-        callback = productId;
-        productId = this.productID;
-      }
-    }
-
-    return this.get(['products', productId, 'book'], { qs: args }, callback);
-  }
-
   cancelAllOrders(args = {}, callback) {
     if (!callback && typeof args === 'function') {
       callback = args;
@@ -189,7 +171,7 @@ class AuthenticatedClient extends PublicClient {
     }.call(this);
 
     if (callback) {
-      p.catch(callback);
+      p.catch(e => setImmediate(() => callback(e)));
     }
 
     return p;
--- a/lib/clients/authenticated.js
+++ b/lib/clients/authenticated.js
@@ -3,8 +3,14 @@ const { signRequest } = require('../../lib/request_signer');
 const PublicClient = require('./public.js');
 
 class AuthenticatedClient extends PublicClient {
-  constructor(key, b64secret, passphrase, apiURI) {
-    super('', apiURI);
+  constructor(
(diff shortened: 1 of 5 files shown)


lib/clients/authenticated.js
```diff
@@ -163,21 +185,14 @@ class AuthenticatedClient extends PublicClient {
...
-        .catch(err => {
-          if (callback) {
-            callback(err);
-          }
-          throw err;
         });
     }.call(this);
 
     if (callback) {
-      p.catch(() => {});
```
nikulis: These conditions ensure that a promise is returned to the user if and only if they did not provide a callback, enforcing the mutually exclusive use of the callback API or promise API per method call. If the user is not dealing with promises, we need to add a blank `catch` handler on the (now internal-only) promise so that any rejected promise will not cause an `UnhandledPromiseRejectionWarning`.

While I like the aesthetics of `p.catch(callback)` much better, it runs into a potential problem with inversion of control; if the user's `callback` `throw`s, then it will either fail silently (on node < 6.6.0), cause an `UnhandledPromiseRejectionWarning` (currently, which is difficult to meaningfully trace during debugging), or terminate the node process (in future node).

nikulis:
Just remembered about #46, which is quite relevant here as well.

lib/clients/authenticated.js
```diff
@@ -3,8 +3,14 @@ const { signRequest } = require('../../lib/request_signer');
...
 
 class AuthenticatedClient extends PublicClient {
-  constructor(key, b64secret, passphrase, apiURI) {
-    super('', apiURI);
+  constructor(
+    key,
+    b64secret,
+    passphrase,
+    apiURI = 'https://api.gdax.com',
+    productID = null
```
fb55: The intention behind this is to make sure that all calls that can change things have an explicit product set. Users shouldn't be trading BTC-USD because they forgot to include the product they actually wanted to trade in the constructor. This shouldn't be a concern for the public client and I would opt to leave it as it is for backwards compatibility.

lib/clients/authenticated.js
fb55: Great comment. I'll wrap this in a `setImmediate`. In general I would opt for always returning a promise and assigning listeners to it in case there is a callback.

lib/clients/authenticated.js
nikulis: Ahh that's a very good point! Gotcha.

In that spirit, should the (authed) constructor throw an error if a `productID` is not explicitly provided?

fb55:
Reopening as this contains some changes I‘d still line to incorporate

rmm5t:
> Reopening as this contains some changes I‘d still line to incorporate

As far ass I can tell, the changes in this PR are now outdated/obsolete in almost every way. 

**What specific changes did you still want to incorporate from here? I'm happy to help extract things out into smaller more-digestible PRs**, but I don't think this PR should remain open. New issues should be created that reference aspects of the desired changes in PR.

lib/clients/authenticated.js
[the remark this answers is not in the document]
jkuchar: Seem good to me!

jkuchar:
I'm missing productId in authenticated API client. (there is no way how to list all fills from other markets then btc-usd in particualar)

rmm5t:
@jkuchar Fills behave a bit differently. In the GDAX API, they don't stem off of the `/products` path like tickers, orderbooks, or trades do, and [they take a different signature](https://docs.gdax.com/#fills) whereby you can filter the fills by `product_id` as a  query string argument. 

For example:

```js
client.getFills({ product_id: 'LTC-USD' }).then(console.log);
```

---

Aside, this PR (#83) is effectively obsolete now. See #178 and #148.

```
npm install gdax-node`
```

jkuchar:
Cool, thanks!

fb55:
Opened #236 for the last change that I wanted to incorporate.

PR #84, opened 2017-07-26
OrderbookSync uses authenticated websocket

jrdi:
It will be interesting to have the ability to listen to authenticated WebSocket when authenticated client is provided. What do you guys think?

### DIFF
--- a/lib/orderbook_sync.js
+++ b/lib/orderbook_sync.js
@@ -10,7 +10,16 @@ class OrderbookSync extends WebsocketClient {
     websocketURI = 'wss://ws-feed.gdax.com',
     authenticatedClient
   ) {
-    super(productIDs, websocketURI);
+    let auth = null;
+    if (authenticatedClient) {
+      auth = {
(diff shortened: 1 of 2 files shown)


fb55:
The PR looks good, just trying to understand the intention: What can be gained from subscribing to the authenticated feed here? Messages would have user- & profile IDs for the user, but there is no place where these are used right now.

jrdi:
@fb55 There is no way to consume "signed" messages within orderbook but you can do it "outside" without the need to create a new websocket.

```javascript
const client = new gdax.AuthenticatedClient(key, secret, passphrase);
const orderbook = new gdax.OrderbookSync(['ETH-EUR'], 'https://api.gdax.com', 'wss://ws-feed.gdax.com', client);

orderbook.on('message', function onMessage(data) {
  // Your code here
});
```

jrdi:
Ping!

fb55:
Sorry for the delay — and thanks for the patch!

PR #87, opened 2017-08-02
Fix infinite request loop on empty data when calling getProductTradeStream()

mistlight:
If tradesTo function always evaluates to false, getProductTradeStream()
will not stop calling the gdax API even if the API returns empty data.


lib/clients/public.js
```diff
@@ -181,6 +181,11 @@ class PublicClient {
           stream.push(data[i]);
         }
 
+        if (data.length === 0) {
```
fb55: Please combine with with the check below (`!loop || data.length === 0`).

tests/mocks/pubclient_stream_trades_function.json
```diff
@@ -1 +1 @@
...0000","side":"sell"},{"time":"2016-05-12T15:21:35.990349Z","trade_id":8408596,"price":"453.38000000","size":"0.03690000","side":"sell"},{"time":"2016-05-12T15:21:35.990059Z","trade_id":8408595,"price":"453.37000000","size":"0.03690000","side":"sell"},{"time":"2016-05-12T15:21:01.234086Z","trade_id":8408594,"price":"453.36000000","size":"0.64393000","side":"buy"},{"time":"2016-05-12T15:21:01.233869Z","trade_id":8408593,"price":"453.36000000","size":"0.01000000","side":"buy"},{"time":"2016-05-12T15:21:01.233699Z","trade_id":8408592,"price":"453.37000000","size":"0.04000000","side":"buy"},{"time":"2016-05-12T15:21:01.233505Z","trade_id":8408591,"price":"453.37000000","size":"0.01000000","side":"buy"},{"time":"2016-05-12T15:21:01.233336Z","trade_id":8408590,"price":"453.38000000","size":"0.04000000","side":"buy"},{"time":"2016-05-12T15:21:01.233146Z","trade_id":8408589,"price":"453.39000000","size":"0.04000000","side":"buy"},{"time":"2016-05-12T15:21:01.23292Z","trade_id":8408588,"price":"453.40000000","size":"0.02400000","side":"buy"},{"time":"2016-05-12T15:21:01.23272Z","trade_id":8408587,"price":"453.40000000","size":"0.01000000","side":"buy"},{"time":"2016-05-12T15:21:01.232543Z","
```
fb55: What changed here?

lib/clients/public.js
mistlight: Done

tests/mocks/pubclient_stream_trades_function.json
mistlight: Added an return blank endpoint, to simulate a call that returns nothing. The only change is the addition of:

```{"scope":"https://api.gdax.com:443","method":"GET","path":"/products/BTC-USD/trades?before=8409514&after=8409615&limit=100","body":"","status":200,"response":[]}``` 
endpoint which is called by getProductTradeStream() as the next endpoint to call when the first one was consumed.

fb55:
Awesome change, thanks a lot @mistlight!

PR #110, opened 2017-09-10
ability to choose which channel to subscribe for

373mich:


runia1:
This looks like a good PR to me.  I noticed the change in the api docs and was wondering if anyone over here had tackled channels yet.  Will use this PR until it gets merged.

fb55:
Tests are currently failing. In general this shouldn't be a breaking change, so ideally you'd add the channels to the end of the argument list, defaulting to `full`

tarrupe:
with this PR, now i can change channels, Thanks! 
But if i use channels with auth i recieve this error:
{"type":"error","message":"Sorry, you could not be authenticated: Bad Request","reason":"invalid signature"}

does anyone else have the same error?

fb55:
@tarrupe The endpoint you need to sign requests for has changed, so if you want to consume the authenticated feed you'll also need to update https://github.com/coinbase/gdax-node/blob/86c298715a8140c466ff91b7cbb24d42d46860a4/lib/clients/websocket.js#L66

to `/users/self/verify`.

runia1:
@fb55 I agree about new args going on the end.  How hard would it be to commandeer this PR, make that change, add a little write up in the readme and merge it?  

I'm guessing the ws feed was changed to "channel" based in order to try and reduce load on your servers.  Seems like it would be top priority for you guys to get this PR or one like it committed here?

fb55:
@runia1 Feel free to take it over :)

The main intention behind channels was to simplify syncing logic and reduce bandwidth requirements for clients, plus enable future extensions. The website as the probably most bandwidth-constrained client already made the switch, getting this out of the door would be great, but sadly isn't a huge priority.

mahn00b:
Hey guys so i've recently been using the api and found that i've been having a similar problem changing channels. I added a parameter to the websocketClient constructor that accepts an array of channel names, if null then will set channel to the default "heartbeat," and saves to local variable called channels. Then I set the subscription request parameter in onOpen to reflect this parameter. It worked when I tested it and switched to "ticker". Below is the file attached with my corrections.

[websocket.txt](https://github.com/coinbase/gdax-node/files/1349518/websocket.txt)
``var client = gdax.WebsocketClient(["LTC-USD"], null, null, ["ticker"]); //works for me

runia1:
@fb55 sorry I have been away for a bit, we bought a house so I've been super busy, otherwise I would have, bit it looks like channels got added in 0.4.4!  Awesome :)

tianheng:
Sorry to disturb here. I don't really know Node.js, I am writing a websocket in Java. I can successfully subscribe any channels except level2. Do you have similar issues in Node.js? I didn't see any tests around level2. The example orderbook_sync.ts is using full channel I think. Correct me if I am wrong. Really appreciate your help.

rmm5t:
This looks like it was added as part of b703f645fa181574e5fa99ac9595f1f04c76fddd

This PR should be safe to close.

aossikine:
Is there any ETA on the merge conflicts and closing of this PR? It looks pretty straightforward and is much needed.

fb55:
Thanks @373mich and @rmm5t

PR #112, opened 2017-09-18
fix(asks): They were in wrong order.

michaelbpaulson:
Asks in 0.4.2 were using `each` but in this version they are using `reach`.  Thus causing them to be backwards.


michaelbpaulson:
A counter fix would be to insert with a different sorter. `b > a` instead of the current `a > b`

fb55:
Thanks a lot @michaelbpaulson!

PR #121, opened 2017-10-10
Updated typing for PublicClient constrctor

MarosPixel:
Optional parameter


fb55:
Nice find, thanks @MarosPixel!

PR #123, opened 2017-10-14
Added proxy support for Websocket

jjanczyk:
Simplistic support for `Websocket` connections via proxy server. Server address & port are taken from `https_proxy` or `http_proxy` environment variables. Not thoroughly tested yet it works via my corporate proxy. Inactive and transparent if no environment is set.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,4 +4,4 @@
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


fb55:
Thanks for the patch! Being able to pass an `agent` property to the websocket constructor would be the much easier and smaller change. Happy to accept a patch that implements that, closing this for now.

PR #126, opened 2017-10-19
Configured array formatting in request extra parameters to 'repeat'

jjanczyk:
This way it is possible to make calls like:

  AuthenticatedClient.getOrders({ status: ['open', 'done'] });

Previously it caused signature errors as arrays were formatted
with indices.


fb55:
Thanks a lot @jjanczyk!

PR #138, opened 2017-11-10
Include response in promise rejections

CjS77:
The response is currently only included in callbacks and dropped in Promise requests.
This PR attaches the response to the error object in rejected promises.


fb55:
Nice one Cayle!

PR #139, opened 2017-11-12
Update README.md

boshdmg:
Clearer explanation of how to use getOrders function


README.md
```diff
@@ -327,7 +327,9 @@ authedClient.cancelAllOrders({product_id: 'BTC-USD'}, callback);
 ```js
 authedClient.getOrders(callback);
 // For pagination, you can include extra page arguments
-authedClient.getOrders({'after': 3000}, callback);
```
fb55: `after` should still be part of it.

PR #143, opened 2017-11-17
take an initial subscribe message instead of productIDs

jflatow:


fb55:
Interesting idea, but that API would be a bit too verbose and break backwards compatibility :/

PR #146, opened 2017-11-22
Adds getPaymentMethods() to authedClient

charltoons:
Needed support for the `payment-methods` API, so I added it.


sagivo:
(dup)
lol, just added similar request - https://github.com/coinbase/gdax-node/pull/171

fb55:
Pretty funny that this is an exact replica of #171. @charltoons thanks a lot for your patch, landed the other PR (which is completely equivalent).

PR #149, opened 2017-11-27
Expose AuthenticatedClient#placeOrder() function

### LINKED ISSUE #98: Why not to expose placeOrder function?
I think it's good to make public `placeOrder` and just set `side` as sell or buy
```js
  _placeOrder(params, callback) {
    let requiredParams = ['side', 'product_id'];
    // ....
    return this.post(['orders'], { body: params }, callback);
  }

  buy(params, callback) {
    params.side = 'buy';
    return this._placeOrder(params, callback);
  }

  sell(params, callback) {
    params.side = 'sell';
    return this._placeOrder(params, callback);
  }
```

In the meantime, I have to do something like `placeOrder` in my code to work around it...

```js
function setOrder(params) {
  params.product_id = params.product_id || 'BTC-USD';
  debug('executing GDAX order: %o', params);
  
  if(params.side === 'buy') {
    return authedClient().buy(params);
  } else if (params.side === 'sell') {
    return authedClient().sell(params);
  } else {
    console.warn(`Invalid order side ${params.side}`);
  }
}
```

rmm5t:
Closes #98

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,6 +296,19 @@ const sellParams = {
 authedClient.sell(sellParams, callback);
 ```
 
+* [`placeOrder`](https://docs.gdax.com/#place-a-new-order)
+
+```js
+// Buy 1 LTC @ 75 USD
+const params = {
+  'side': 'buy',
+  'price': '75.00', // USD
+  'size': '1', // LTC
+  'product_id': 'LTC-USD',
+};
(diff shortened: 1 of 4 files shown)


fb55:
Thanks a lot @rmm5t!

PR #150, opened 2017-11-30
Fix heartbeat channel subscriptions

### LINKED ISSUE #113: GDAX Heartbeat Channel
Hi there.  I'm looking at the documentation and trying to subscribe to the heartbeat channel, with this code:
```

var websocket = new Gdax.WebsocketClient(
    ['BTC-USD'],
    'wss://ws-feed.gdax.com',
    {
        key: API_KEY,
        secret: API_SECRET,
        passphrase: API_PASSPHRASE,
    },
    { heartbeat: true }
)
```
```

  webSocket.on('message', data => {
    console.log(data);
  });

```
But I keep receiving the following error, despite passing BTC-USD
//error
`{ type: 'error',message: 'Failed to subscribe',reason: 'You need to specify at least one product ID for channel heartbeat'
`

I'm guessing this is more to do with my ignorance on how the code works than anything else.  Is there a more appropriate, centralized place to put these comments?

rmm5t:
- Automatically adds the `heartbeat` channel if not already subscribed
- Removes redundant `keepalive` ping on the socket

Fixes #113

### DIFF
--- a/lib/clients/websocket.js
+++ b/lib/clients/websocket.js
@@ -14,14 +14,16 @@ class WebsocketClient extends EventEmitter {
     productIDs,
     websocketURI = 'wss://ws-feed.gdax.com',
     auth = null,
-    { heartbeat = false, channels = null } = {}
+    { channels = null } = {}
   ) {
     super();
     this.productIDs = Utils.determineProductIDs(productIDs);
     this.websocketURI = websocketURI;
-    this.channels = channels;
     this.auth = Utils.checkAuth(auth);
-    this.heartbeat = heartbeat;
+    this.channels = channels || ['full'];
+    if (!this.channels.includes('heartbeat')) {
+      this.channels.push('heartbeat');
+    }
     this.connect();
   }
 
@@ -55,11 +57,8 @@ class WebsocketClient extends EventEmitter {
     const subscribeMessage = {
       type: 'subscribe',
       product_ids: this.productIDs,
+      channels: this.channels,
     };
-
-    if (this.channels) {
-      subscribeMessage.channels = this.channels;
-    }
 
     // Add Signature
     if (this.auth.secret) {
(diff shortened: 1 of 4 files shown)


lib/clients/websocket.js
```diff
@@ -24,15 +24,16 @@ class WebsocketClient extends EventEmitter {
         'Invalid or incomplete authentication credentials. You should either provide all of the secret, key and passphrase fields, or leave auth null'
       );
     }
-    this.channels = channels;
     this.auth = auth || {};
-    this.heartbeat = heartbeat;
+    this.channels = channels || ['full'];
+    if (!this.channels.includes('heartbeat')) {
+      this.channels.push('heartbeat');
```
fb55: Heartbeats are pretty noisy, so there should be a way to disable them to save bandwidth.

fb55:
I like the change, just one comment

lib/clients/websocket.js
rmm5t: > Heartbeats are pretty noisy...

🤔 Not compared the `full` or `ticker` or `matches` or `level2` channel.  The only channel that is less noisy than `heartbeat` is the `user` channel.

`heartbeat` is mild compared to anything else, and it comes with other very useful benefits (keeps the connection open, helps address lost sequences, etc).

Also, the longer-term intent of this automatic `heartbeat` addition is to eventually allow the `WebsocketClient` to automatically monitor the heartbeat, reconnect as necessary, and potentially fire an event when a missed sequence is recognized. i.e. The `heartbeat` does not need to necessarily propagate to the caller.

(I'll also rebase after the new conflicts.)

rmm5t:
Rebased. Let me know if the automatic `heartbeat` channel is a show stopper.  I'm in favor of keeping the automatic `heartbeat`. Without it, I suspect we'll need to add the keepalive socket ping back in (i.e all the deleted code from this pull-request and all the extra liability of 100+ lines of code that comes with it).

rmm5t:
Rebased again to resolve new conflicts.

fb55:
Thanks a lot @rmm5t!

PR #151, opened 2017-11-30
Refactor auth object to { key, secret, passphrase }

rmm5t:
This more consistent usage allows an `AuthenticatedClient` instance to be used
in place of an `auth` object thereby simplifying codebase and general usage patterns.

`AuthenticatedClient` used to have a `b64secret` property, but all auth object credentials expected a `secret` property. By making the two consistent, we gain better consistency, and the codebase can be simplified a bit.

**Other considerations**
This will likely need to be rebased after #150 is merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,13 +219,13 @@ the API URI (defaults to `https://api.gdax.com`).
 
 ```js
 const key = 'your_api_key';
-const b64secret = 'your_b64_secret';
+const secret = 'your_b64_secret';
 const passphrase = 'your_passphrase';
 
 const apiURI = 'https://api.gdax.com';
 const sandboxURI = 'https://api-public.sandbox.gdax.com';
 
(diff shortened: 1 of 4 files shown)


rmm5t:
I should also mention that if this refactor is accepted, I'd also like to submit a small change such that the `OrderbookSync` doesn't necessarily ask for an `AuthenticatedClient`, but instead just an optional `auth` object (i.e. `{ key, secret, passphrase }`).  Afterall, the `authenticatedClient` config might conflict with the other `OrderbookSync` arguments, like `apiURI`.

(This would still remain backwards compatible, because the property signatures would still behave the same way, but it would free up the library as a whole to move towards a better consistency)

fb55:
Great change. Also in favor of the proposed extension, no reason to need an authenticated client here.

Issue #153, opened 2017-12-02
promise method from sample code doesn't work


markoshust:
using the below script:

```
const Gdax = require('gdax');
const publicClient = new Gdax.PublicClient();

publicClient
  .getProducts()
  .then(data => {
    console.log('data', data);
    // work with data
  })
  .catch(error => {
    console.log('error', error);
    // handle the error
  });
```

expected: data and error logged to console
actual: 
```
./node_modules/gdax/lib/clients/public.js:57
      throw "Must supply a callback."
      ^
```

markoshust:
It appears the npm version of the module is out of date. i'm assuming this is not auto-published, so recommending to add a note to the readme about which version to use, as i was really wondering why the code wasn't working.

rmm5t:
@markoshust Would you mind closing this issue in favor of consolidating the discussion under #148? There's a deeper discussion going on there.

Issue #148, opened 2017-11-27
v0.5.0 is missing from npm


rmm5t:
* [GitHub Releases](https://github.com/coinbase/gdax-node/releases) shows v0.5.0.
* [NPM](https://www.npmjs.com/package/gdax) only shows v0.4.2.

**Will a maintainer please `npm publish` v0.5.0?**

----

Known workarounds in the meantime:

`npm install coinbase/gdax-node`
or
`yarn add coinbase/gdax-node`

----

Overall, there is a lack of discipline on this project with regards to releases:

* [Haphazard and missing release notes](https://github.com/coinbase/gdax-node/releases)
* [Missing tags](https://github.com/coinbase/gdax-node/releases) for releases that are [on NPM](https://www.npmjs.com/package/gdax)
* [Missing NPM releases for tags that are on the project](https://github.com/coinbase/gdax-node/issues/148).
* [Violations of semver](https://github.com/coinbase/gdax-node/commit/b708da1bd89eb22dcf229bdd5687032f9463de6e#comments)
* Automation-wise, it's evident that neither `yarn version` nor `npm version` and `npm publish` are not used properly (nor at the same times) when actually building new releases; otherwise, the tagging and NPM publishing history wouldn't be the mess that it currently is.

----

UPDATE: There's now a v0.5.1 release on NPM (skipping v0.5.0), but this release violated semver and there's no associated tag on the repository. I think this particular issue should remain open until we fix the conventions around release management (hopefully as part of a v0.6.0 release).

BrandonCopley:
I can help update the npm module, it's a really easy push. It wouldn't be secure, but maybe someone on Gdax could push v0.5.0 up to NPM

rmm5t:
@BrandonCopley I've sent multiple emails to @mihar about volunteering to help maintain this project. No response in 2 weeks. 😞 

If anyone else has the keys and would like me to help maintain this project, I'm making myself available. Please feel free to review my open pull-requests on this project. Feel free to review my [GitHub open-source history](https://github.com/rmm5t) and my [StackOverflow profile](https://github.com/rmm5t) for credentials.

Here's a quick summary of what I think should be a priority for gdax-node, and I'm open to hearing other people's thoughts on the matter:

- Improve process for stable releases (including versioning, publishing, and documenting).
- Move towards a more stable v1.0 release (along with careful considerations of backwards-incompatible changes and interim releases leading up to a v1.0 release).
- Improve test coverage.
[...]

PR #154, opened 2017-12-02
updated README with note on npm version usage

markoshust:
closes https://github.com/coinbase/gdax-node/issues/153

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,10 +25,10 @@ the MIT License – please take the following message to heart:*
 ## Installation
 
 ```bash
-# from npm
+# from npm (may be outdated)
 npm install gdax
 
-# from GitHub
+# from GitHub (most recent code)
 npm install coinbase/gdax-node
 ```
 


rmm5t:
The is not the appropriate change to make for the problem of the maintainers failing to publish to NPM.

fb55:
Agreed. We shouldn’t even mention the Github path in the first place. Right now we don’t want to publish a new release as master contains a lot of breaking changes, and a minor version bump, even if it is semver compatible, could be surprising to users. At the same time, the library isn’t in a state where we are happy to publish a 1.0 either. I’m hoping to get there in the next weeks.

rmm5t:
> Right now we don’t want to publish a new release as master contains a lot of breaking changes, and a minor version bump, even if it is semver compatible, could be surprising to users.

@fb55 I couldn't disagree more here, because hat violates the entire purpose of semver when the major version is still `0`.  A minor version bump when the major version is still `0` implies breaking compatibility. Right now, the README documents things that are not on NPM, and that's obviously very confusing to newcomers (according to the onslaught of unnecessary Issues on this project)

It's best to continue to bump 0.x versions to NPM. Otherwise, the amount of work required for someone to upgrade from v0.4 to v1.x would be too large and daunting. It's best to have lots of little releases (even if some are backward incompatible), where someone can walk along a corresponding changelog and upgrade incrementally as as necessary.

markoshust:
I don’t know why this PR was closed. This is a stop gap to please your user
base so they don’t get pissed off when your npm package doesn’t work as the
readme suggests it should.

Either a new 0.X version should be published as suggested, or this should
be tagged as 1.0.0 if there are no future breaks expected, even if it
doesn’t completely work. Fixes can come in 1.0.1 or 1.1.0. If you are
waiting until everything is perfect, you will be waiting a lot longer than
a few weeks.

rmm5t:
@markoshust The PR was likely closed, because the suggested change isn't the appropriate long-term change to make. I'd argue it isn't the best short-term solution either.

> ... a new 0.X version should be published as suggested

Agreed. IMO, the appropriate approach is to resolve Issue #148 and get new versions on NPM.  At the end of the day, the tagged releases in GitHub should match what's on NPM.

PR #160, opened 2017-12-14
Fix WebsocketClient typescript definition

iffy:


GreyEcologist:
Is this why wires are taking forever to appear on GDAX? If not is there a ticket or PR someone is working and needs help with. With the support team being non-existent, ill-equipped and/or incompetent, I'd love to resolve the issue myself if I can. So, if someone can point me in the right direction of what the hell is going with bank wires, I would really appreciate it.

iffy:
@GreyEcologist This change has nothing to do with delays; just with the TypeScript definition for the Websocket-related code.

fb55:
Thanks a lot @iffy!

Issue #169, opened 2017-12-20
TypeScript definition for WebsocketClient doesn't match the actual arguments


ncallaway:
The exported TypeScript definition for the constructor of WebsocketClient only declares a single argument of `productIds`. The actual constructor takes several arguments, including `productIDs`, `websocketURI`, `auth`, and an object describing the channels to subscribe to.

**TypeScript Definition**
https://github.com/coinbase/gdax-node/blob/9a59db269857d10cb449c44cd6a0f51c630abd5c/index.d.ts#L243

```
    export class WebsocketClient {
        constructor(productIds: string[]);

        on(event: 'message', eventHandler: (data) => void);
        on(event: 'error', eventHandler: (err) => void);
        on(event: 'open', eventHandler: () => void);
        on(event: 'close', eventHandler: () => void);
    }
```

**Websocket Constructor**

https://github.com/coinbase/gdax-node/blob/9a59db269857d10cb449c44cd6a0f51c630abd5c/lib/clients/websocket.js#L13

 ```
 constructor(
    productIDs,
    websocketURI = 'wss://ws-feed.gdax.com',
    auth = null,
    { heartbeat = false, channels = null } = {}
  )
```

**Expected Behavior**
When using TypeScript, we should be able to construct a WebSocket client that can pass in the necessary parameters, including `websocketURI`, `auth`, `heartbeat` and `channels`.

Example:
```
const establishedSocket = new Gdax.WebsocketClient(
      ["BTC-USD"],
      api.wsEndpoint,
      { key: sandbox.key, secret: sandbox.secret, passphrase: sandbox.passphrase },
      { channels: ["ticker", "user"] }
    );
```

**Actual Behavior**
When using TypeScript, if any parameters aside from `productIDs` are provided, a TypeScript compile error occurs.

Error:
`TS2554: Expected 1 arguments, but got 4.`

**Proposed Fix**
Update the TypeScript definition to match the actual constructor of the WebsocketClient. I'll have a pull request available with this proposed fix shortly.

ncallaway:
This issue would also be addressed by pull request #160.

fb55:
Fixed in #160

PR #170, opened 2017-12-21
Update TypeScript declarations for WebsocketClient

ncallaway:
This updates the TypeScript declarations for the WebsocketClient to match the constructor that's available in javascript.

This is a proposed resolution to Issue #169.

I didn't see any existing mechanism for testing these type declarations, so I didn't add any tests for this. If there is a mechanism for testing type declarations, please let me know and I'll add tests.

All arguments were named to match the names in the javascript WebsocketClient constructor. The only exception is the `channels` argument, which is not specifically named in the javascript as it's an anonymous destructured object. I simply went with the name that seemed to best describe the object itself.


ncallaway:
I didn't see the other pull request that resolves this issue by iffy here at #160 . If that pull request is accepted, please feel free to close this pull request.

index.d.ts
```diff
@@ -239,8 +239,24 @@ declare module 'gdax' {
         getTrailingVolume(): Promise<any>;
     }
 
+    export interface WebsocketAuthentiation {
```
wingrunr21: You are missing a `c` in `Authentication`

wingrunr21:
Thanks for this! I was just getting ready to fork/PR to fix this.

ncallaway:
Good catch on the spelling! At least it was consistent ;)

fb55:
@mcallaway Sorry, didn’t see this before merging the other one. This seems a bit more complete, would be happy to merge after a rebase!

index.d.ts
```diff
@@ -239,8 +239,24 @@ declare module 'gdax' {
...
+        secret: string,
+        passphrase: string
+    }
+
+    export interface WebsocketChannels {
+        heartbeat?: boolean,
+        channels?: string[],
+    }
+
     export class WebsocketClient {
```
itajaja: shouldn't this extend `EventEmitter`? `@types/node` would have to be installed

PR #177, opened 2017-12-25
Simplify OrderbookSync constructor signature

rmm5t:
This is a direct follow-up to https://github.com/coinbase/gdax-node/pull/151#issuecomment-348232843

- Adds a `checkAuth()`  utility function to ensure an `auth` object contains `key`, `secret`, and `passphrase`
- Simplifies `OrderbookSync` constructor to take auth object instead of a full `AuthenticatedClient` instance

This is still backwards compatible with the old constructor signature, because `AuthenticatedClient` has the same `key`, `secret`, and `passphrase` properties.

/cc @fb55

### DIFF
--- a/lib/orderbook_sync.js
+++ b/lib/orderbook_sync.js
@@ -1,6 +1,8 @@
 const WebsocketClient = require('./clients/websocket.js');
+const AuthenticatedClient = require('./clients/authenticated.js');
 const PublicClient = require('./clients/public.js');
 const Orderbook = require('./orderbook.js');
+const Utils = require('./utilities.js');
 
 // Orderbook syncing
 class OrderbookSync extends WebsocketClient {
@@ -8,17 +10,26 @@ class OrderbookSync extends WebsocketClient {
     productIDs,
     apiURI = 'https://api.gdax.com',
     websocketURI = 'wss://ws-feed.gdax.com',
-    authenticatedClient = null,
+    auth = null,
     { heartbeat = false } = {}
   ) {
-    super(productIDs, websocketURI, authenticatedClient, { heartbeat });
+    super(productIDs, websocketURI, auth, { heartbeat });
     this.apiURI = apiURI;
-    this.authenticatedClient = authenticatedClient;
+    this.auth = Utils.checkAuth(auth);
 
     this._queues = {}; // []
     this._sequences = {}; // -1
     this._public_clients = {};
     this.books = {};
+
+    if (this.auth.secret) {
+      this._authenticatedClient = new AuthenticatedClient(
+        this.auth.key,
+        this.auth.secret,
+        this.auth.passphrase,
+        this.apiURI
+      );
+    }
 
     this.productIDs.forEach(productID => {
       this._queues[productID] = [];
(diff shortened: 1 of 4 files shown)


tests/orderbook_sync.spec.js
```diff
@@ -147,18 +141,17 @@ suite('OrderbookSync', () => {
     });
   });
 
-  test('emits an error event on error (with AuthenticatedClient)', done => {
```
fb55: Could one of these still use an authenticated client, to ensure we don't break backwards compatibility?

lib/orderbook_sync.js
```diff
@@ -1,25 +1,36 @@
...
-    this.authenticatedClient = authenticatedClient;
+    this.auth = Utils.checkAuth(auth);
 
     this._queues = {}; // []
     this._sequences = {}; // -1
     this._public_clients = {};
     this.books = {};
 
+    if (this.auth.secret) {
+      this._authenticatedClient = new AuthenticatedClient(
```
fb55: Can this default to `auth` if it already is an instance of `AuthenticatedClient`?

tests/orderbook_sync.spec.js
rmm5t: Personally, I don't think we should have tests for that, because we're moving forward with the API. The benefit of previously changing the `AuthenticatedClient` properties and the backwards compatibility that comes with it is _**just a nice and elegant side effect**_ . Backwards compatibility guarantees can be good sometimes, but they can also hold things back long-term. 

Nonetheless, with that off my chest, **I'm happy to add an extra test with `AuthenticatedClient` if would make you feel more comfortable.** I'm just not sure this is the correct place to test that `AuthenticatedClient` quacks like an `auth` payload object of `{ key, secret, passphrase }`.

lib/orderbook_sync.js
rmm5t: I vote note. For the same reasons as above. We should not be making side-effect assumptions about the passed object, aside from the fact that it quacks like a `{ key, secret, passphrase }` object. Otherwise, the project opens itself to a world of forwards compatibility issues and limitations down the road. 

i.e. "Optimizations" like this only yield future maintenance problems. There's almost zero overhead to having an extra `AuthenticatedClient` instance. 

e.g. If we did reuse the passed-in `AuthenticationClient`, the `apiURI` argument passed to the `OrderbookSync` constructor would then be ignored, and you'd end up with unintended consequences of talking to the wrong endpoint.

tests/orderbook_sync.spec.js
rmm5t: An extra test using an instance of `AuthenticatedClient` was just added via 09be1e8.

fb55:
Awesome, thanks a lot!

blair:
It appears to me that this commit goes in the opposite direction of dependency injection: https://en.wikipedia.org/wiki/Dependency_injection . Normally, one passes in all the objects that an object needs in its constructor and constructors don't construct other objects. It's ok to have to build an AuthenticatedClient and then pass it to a OrderbookSync even if the code is a little longer.

This principle also makes it easier to pass in mock objects if one wanted to. Having a constructor construct objects itself removes this possibility.

If you're concerned about the two URIs being out of sync, then just drop the URI argument.

Unless there's something I'm missing, I suggest reverting this commit and only taking a PublicClient instance (which should have AuthenticatedClient as a sublcass, but that doesn't look like it was fixed until a week ago in https://github.com/coinbase/gdax-node/commit/60f8014007b15451c816e1714fdd2704c92e30b9 ). So wait until gdax-tt upgrades to the latest gdax.

PR #178, opened 2017-12-26
Normalize PublicClient with AuthenticatedClient 

### LINKED ISSUE #46: Authenticated getProductOrderBook overridden

### LINKED ISSUE #117: AuthenticatedClient doesn't allow specification of a productID for inherited PublicClient methods
The [documentation](https://github.com/coinbase/gdax-node/blob/master/README.md#the-authenticated-api-client)  says that "AuthenticatedClient inherits all of the API methods from PublicClient, so if you're hitting both public and private API endpoints you only need to create a single client."

I'm having difficulty using the PublicClient methods via the AuthenticatedClient to e.g. get trade history against a currency pair other than USD-BTC. 

From my cursory investigation, the problem seems to be that the PublicClient constructor accepts a productID, but the AuthenticatedClient does not. In fact, the AuthenticatedClient inherits from PublicClient and explicitly passes an empty string as a productID to super:

https://github.com/coinbase/gdax-node/blob/6fd3d51d4007d19566335cdbe0bedde02e0998ee/lib/clients/authenticated.js#L5-L7

https://github.com/coinbase/gdax-node/blob/6fd3d51d4007d19566335cdbe0bedde02e0998ee/lib/clients/public.js#L4-L6

Thus when calling any of the inherited PublicClient methods like getProductTicker, this.product_id would seem to incorrectly be set to an empty string (or defaulted to USD-BTC): 

https://github.com/coinbase/gdax-node/blob/6fd3d51d4007d19566335cdbe0bedde02e0998ee/lib/clients/public.js#L107-L109

What's going on here? If I'm missing something (which seems most likely), please let me know what I'm missing.

Thanks.

### LINKED ISSUE #158: Authed client can't get product ticker
The docs clearly state

> AuthenticatedClient inherits all of the API methods from PublicClient, so if you're hitting both public and private API endpoints you only need to create a single client.

But if I run `const tick = await authedClient.getProductTicker();`, I get `{ "message": "NotFound" }`.

The authedClient is created correctly because I can perform authedClient.buy() just fine.

rmm5t:
## What does it do?

- **Deprecates `productID` argument in `PublicClient` constructor**
- Requires `productID` as first argument to `getProductOrderBook()` and deprecates old usage signature
- Removes redundant `AuthenticatedClient#getProductOrderBook()` override
- Simplifies `OrderbookSync#loadOrderbook()` because both clients now adhere to the same method signature
- Requires `productID` as first argument to `getProductTicker()` and deprecates old usage signature
- Requires `productID` as first argument to `getProductTrades()` and deprecates old usage signature
- Requires `productID` as first argument to `getProductTradeStream()` and deprecates old usage signature
- Requires `productID` as first argument to `getProductHistoricRates()` and deprecates old usage signature
- Requires `productID` as first argument to `getProduct24HrStats()` and deprecates old usage signature
- Consolidates deprecation & argument normalization logic in `PublicClient` to later simplify the eventual transition of outright removing the (now unnecessary) `PublicClient#productID` property.
- Updates the TypeScript definitions to match
- Updates the `README` documentation to match
- **Includes several tests that were missing**
- **Includes backwards compatibility tests for all deprecated method signatures**

- **Any use of a deprecated method signature will emit a warning**
- The general idea is that `PublicClient` should no longer maintain state about a particular product ID.
- Any method that can be called on a `PublicClient` instance can now also be called on an `AuthenticatedClient` instance.  In other words, the classes now correctly adhere to the specification in the `README`:
    > _"`AuthenticatedClient` inherits all of the API methods from `PublicClient`, so if you're hitting both public and private API endpoints you only need to create a single client."_

## Related Issues and PRs

- Fixes #46
- Fixes #117
- Fixes #158
- Closes #83 (this PR supersedes)
- Closes #131 (this PR supersedes)

## Communication

- Because this is a fairly large PR, **I recommend code-reviewing the changes for each commit individually instead of reviewing the entire PR at once.** It's likely easier to digest that way.

- **I think it would be prudent to make an interim v0.x release with these changes, thereby documenting and committing to all the deprecations. That would allow for a full cleanup during an eventual v1.0 release where the old method signatures and deprecations are removed, and the new method signature win-out.**

/cc @fb55

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,13 +153,13 @@ publicClient.getProducts(callback);
 publicClient.getProductOrderBook('BTC-USD', callback);
 
 // Get the order book at a specific level of detail.
-publicClient.getProductOrderBook('LTC-USD', {'level': 3}, callback);
+publicClient.getProductOrderBook('LTC-USD', { level: 3 }, callback);
 ```
 
 * [`getProductTicker`](https://docs.gdax.com/#get-product-ticker)
 
 ```js
-publicClient.getProductTicker(callback);
+publicClient.getProductTicker('ETH-USD', callback);
 ```
 
 * [`getProductTrades`](https://docs.gdax.com/#get-trades)
@@ -168,7 +168,7 @@ publicClient.getProductTicker(callback);
 publicClient.getProductTrades(callback);
 
 // To make paginated requests, include page parameters
-publicClient.getProductTrades({'after': 1000}, callback);
+publicClient.getProductTrades({ after: 1000 }, callback);
 ```
 
 * [`getProductTradeStream`](https://docs.gdax.com/#get-trades)
@@ -189,7 +189,7 @@ const trades = publicClient.getProductTradeStream(8408000, trade => Date.parse(t
 publicClient.getProductHistoricRates(callback);
 
 // To include extra parameters:
-publicClient.getProductHistoricRates({'granularity': 3000}, callback);
+publicClient.getProductHistoricRates({ granularity: 3000 }, callback);
 ```
 
 * [`getProduct24HrStats`](https://docs.gdax.com/#get-24hr-stats)
@@ -269,7 +269,7 @@ const accountID = '7d0f7d8e-dd34-4d9c-a846-06f431c381ba';
 authedClient.getAccountHistory(accountID, callback);
 
 // For pagination, you can include extra page arguments
-authedClient.getAccountHistory(accountID, {'before': 3000}, callback);
+authedClient.getAccountHistory(accountID, { before: 3000 }, callback);
 ```
 
 * [`getAccountHolds`](https://docs.gdax.com/#get-holds)
@@ -279,7 +279,7 @@ const accountID = '7d0f7d8e-dd34-4d9c-a846-06f431c381ba';
 authedClient.getAccountHolds(accountID, callback);
 
 // For pagination, you can include extra page arguments
-authedClient.getAccountHolds(accountID, {'before': 3000}, callback);
+authedClient.getAccountHolds(accountID, { before: 3000 }, callback);
 ```
 
 * [`buy`, `sell`](https://docs.gdax.com/#place-a-new-order)
@@ -346,7 +346,7 @@ authedClient.cancelAllOrders({product_id: 'BTC-USD'}, callback);
 ```js
 authedClient.getOrders(callback);
 // For pagination, you can include extra page arguments
-authedClient.getOrders({'after': 3000}, callback);
+authedClient.getOrders({ after: 3000 }, callback);
 ```
 
 * [`getOrder`](https://docs.gdax.com/#get-an-order)
@@ -361,7 +361,7 @@ authedClient.getOrder(orderID, callback);
 ```js
 authedClient.getFills(callback);
 // For pagination, you can include extra page arguments
-authedClient.getFills({'before': 3000}, callback);
+authedClient.getFills({ before: 3000 }, callback);
 ```
 
 * [`getFundings`](https://docs.gdax.com/#list-fundings)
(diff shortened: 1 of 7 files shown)


fb55:
Awesome PR, thanks a lot!

Regarding npm releases: How do you feel about doing 1.0 beta releases?

rmm5t:
> Regarding npm releases: How do you feel about doing 1.0 beta releases?

@fb55 For this project (and almost all open source projects), I'm always going to advocate for more frequent minor releases vs one large major release drop. Here, this means _at least_ an official v0.5 (back) release and a v0.6 release prior to going to v1.0. 

**Reasons:**

1. When true deprecations are introduced to a codebase (as they are in this PR), it's imperative that those deprecations get published out in a minor release as soon as possible. It's best to have the community run into them early and fix the warnings before the deprecation warnings are eventually  removed.
1. When a major version bump happens, the deprecations should be removed, and the codebase made as clean as possible (avoiding the old deprecation conditional checks).
1. Almost no one tests `beta` or `rc1` (release candidate) releases. They are only valuable for really large progjects (operating systems, web frameworks, languages) and/or really popular projects that have large upgrade paths and potentially untested territory. This project is much too small to glean any kind of benefit from a `beta` release.
1. True and more-frequent semver'd releases allow developers to more easily peg themselves to an older version of a library if they don't yet have the time to perform a major upgrade. It not only improves stability amongst the community, but it also instills confidence in the open source library itself and the maintainers themselves. 

In summary, an official `beta` release will yield almost no benefit to this project. Frequent minor releases will get used and allow us to iterate more quickly and with more confidence.

----

Also, with all that said, even with this PR's normalization of `PublicClient` and `AuthenticationClient`, I think there is still room to improve the method signatures through the library.

1. Positional arguments for a library like this that wraps another REST API makes little sense. I think we should move this library towards keyword arguments as much as possible. 
    1. Keyword arguments better document themselves. 
    1. Keyword arguments allow for easier forwards compatibility changes. We can add features without affecting existing codebases.
1. The constructor for `PublicClient` and `AuthenticationClient` still don't match. This is a code-smell from an objected-oriented perspective. Both should use keyword arguments. This would allow you to instantiate either client interchangeably based on a variable pointing at the `class` as long as you expect to use the parent interface. e.g. This would further cleanup the `OrderbookSync` internal implementation.

Another concrete example:

```js
// Existing API signature (positional arguments make things unnecessarily messy)
const ob = new OrderBookSync(['BTC-USD'],  'https://api.gdax.com', 'wss://ws-feed.gdax.com', { key, secret, passphrase })

//  Future potential API signature (far superior with keyword arguments)
const productIDs = ['BTC-USD'];
const auth = { key, secret, passphrase };
const ob = new OrderBookSync({ productIDs, auth }); order doesn't matter, no duplicate defaults
```

**Perhaps, keyword arguments need to wait until a v2.0 release**, _BUT_ these are the kinds of considerations that make more-frequent releases more important, because they allow the community to keep up with rapid development and change. Allow for change and prepare for change, but don't hold too tightly to backwards compatibility forever. 

Forwards compatibility is the more interesting and beneficial design paradigm.

tkriplean:
Thanks @rmm5t!

vansergen:
It seems that https://github.com/coinbase/coinbase-pro-node/pull/362 solves the issues that @rmm5t mentioned
> 1. Positional arguments for a library like this that wraps another REST API makes little sense. I think we should move this library towards keyword arguments as much as possible.
>    
>    1. Keyword arguments better document themselves.
>    2. Keyword arguments allow for easier forwards compatibility changes. We can add features without affecting existing codebases.
> 2. The constructor for `PublicClient` and `AuthenticationClient` still don't match. This is a code-smell from an objected-oriented perspective. Both should use keyword arguments. This would allow you to instantiate either client interchangeably based on a variable pointing at the `class` as long as you expect to use the parent interface. e.g. This would further cleanup the `OrderbookSync` internal implementation.

PR #179, opened 2017-12-26
Add subscribe/unsubscribe functions to websocket.

fuzzyTew:
I think it would be great if the websocket client allowed changing what channels are subscribed without losing the connection.

How does this proposed API look?

### DIFF
--- a/lib/clients/websocket.js
+++ b/lib/clients/websocket.js
@@ -17,13 +17,11 @@ class WebsocketClient extends EventEmitter {
     { channels = null } = {}
   ) {
     super();
-    this.productIDs = Utils.determineProductIDs(productIDs);
+    this.productIDs = new Set(Utils.determineProductIDs(productIDs));
     this.websocketURI = websocketURI;
     this.auth = Utils.checkAuth(auth);
-    this.channels = channels || ['full'];
-    if (!this.channels.includes('heartbeat')) {
-      this.channels.push('heartbeat');
-    }
+    this.channels = new Set(channels || ['full']);
+    this.channels.add('heartbeat');
     this.connect();
   }
 
@@ -48,26 +46,62 @@ class WebsocketClient extends EventEmitter {
     this.socket = null;
   }
 
-  onOpen() {
-    this.emit('open');
-
+  _sendSubscription(
+    type,
+    productIDs = this.productIDs,
+    channels = this.channels
+  ) {
     const subscribeMessage = {
-      type: 'subscribe',
-      product_ids: this.productIDs,
-      channels: this.channels,
+      type: type,
+      channels: [...channels],
     };
+
+    if (productIDs) {
+      subscribeMessage.product_ids = [...productIDs];
+    }
 
     // Add Signature
     if (this.auth.secret) {
-      let sig = signRequest(
-        this.auth,
-        'GET',
-        this.channels ? '/users/self/verify' : '/users/self'
-      );
+      let sig = signRequest(this.auth, 'GET', '/users/self/verify');
       Object.assign(subscribeMessage, sig);
     }
 
     this.socket.send(JSON.stringify(subscribeMessage));
+  }
+
+  subscribeProduct(...productIDs) {
+    productIDs.forEach(p => this.productIDs.add(p));
+
+    this._sendSubscription('subscribe');
+  }
+
+  subscribeChannel(...channels) {
+    channels.forEach(c => this.channels.add(c));
+
+    this._sendSubscription('subscribe');
+  }
+
+  unsubscribeProduct(...productIDs) {
+    productIDs.forEach(p => this.productIDs.delete(p));
+
+    var channels = [...this.channels].map(c => ({
+      name: c,
+      product_ids: productIDs,
+    }));
+
+    this._sendSubscription('unsubscribe', null, channels);
+  }
+
+  unsubscribeChannel(...channels) {
+    channels.forEach(c => this.channels.delete(c));
+
+    this._sendSubscription('unsubscribe', null, channels);
+  }
+
+  onOpen() {
+    this.emit('open');
+
+    this._sendSubscription('subscribe');
   }
 
   onClose() {
--- a/lib/orderbook_sync.js
+++ b/lib/orderbook_sync.js
@@ -31,20 +31,39 @@ class OrderbookSync extends WebsocketClient {
       this._client = new PublicClient(this.apiURI);
     }
 
-    this.productIDs.forEach(productID => {
-      this._queues[productID] = [];
-      this._sequences[productID] = -2;
-      this.books[productID] = new Orderbook();
-    });
+    this.productIDs.forEach(this._newProduct, this);
+  }
+
+  _newProduct(productID) {
+    this._queues[productID] = [];
+    this._sequences[productID] = [];
+    this.books[productID] = new Orderbook();
   }
 
   onMessage(data) {
     data = JSON.parse(data);
     this.emit('message', data);
 
-    const { product_id } = data;
+    const { type, product_id } = data;
 
-    if (this._sequences[product_id] < 0) {
+    if (type === 'subscriptions') {
+      // subscriptions changed -- possible new products
+      var newProducts = new Set();
+
+      data.channels.forEach(c => {
+        if (c.name === 'full') {
+          newProducts = new Set(c.product_ids);
+        }
+      });
+
+      newProducts.forEach(productID => {
+        if (productID in this.books) {
+          return;
+        }
+
+        this._newProduct(productID);
+      });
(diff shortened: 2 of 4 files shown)


fuzzyTew:
To summarize, this patch adds the following functions to WebsocketClient:

- subscribeProduct(..products)
- subscribeChannel(...channels)
- unsubscribeProduct(...products)
- unsubscribeChannel(...channels)

This allows the specified subscriptions to be changed dynamically.  The code maintains the set of channels and products such that for each channel, one is always subscribed to the same set of products as all other channels.

lib/clients/websocket.js
```diff
@@ -17,12 +17,12 @@ class WebsocketClient extends EventEmitter {
...
+    this.productIDs = new Set(Utils.determineProductIDs(productIDs));
     this.websocketURI = websocketURI;
     this.auth = Utils.checkAuth(auth);
-    this.channels = channels || ['full'];
-    if (!this.channels.includes('heartbeat')) {
-      this.channels.push('heartbeat');
+    this.channels = new Set(channels || ['full']);
+    if (!this.channels.has('heartbeat')) {
+      this.channels.add('heartbeat');
     }
```
rmm5t: Good use of `Set` here, but there's no need to check if `channels` already has `heartbeat` before adding it. That's the benefit of using the `Set`. 😃

lib/clients/websocket.js
```diff
@@ -48,28 +48,68 @@ class WebsocketClient extends EventEmitter {
...
       Object.assign(subscribeMessage, sig);
     }
 
     this.socket.send(JSON.stringify(subscribeMessage));
   }
 
+  subscribeProduct(...productIDs) {
+    for (let p of productIDs) {
+      this.productIDs.add(p);
+    }
```
rmm5t: This might be my personal preference, but I much prefer `Array.prototype.forEach()` usage patterns. For example:


```js
productIDs.forEach(p => this.productIDS.add(p);
```

lib/clients/websocket.js
```diff
@@ -48,28 +48,68 @@ class WebsocketClient extends EventEmitter {
...
+
     // Add Signature
     if (this.auth.secret) {
-      let sig = signRequest(
-        this.auth,
-        'GET',
-        this.channels ? '/users/self/verify' : '/users/self'
-      );
+      let sig = signRequest(this.auth, 'GET', '/users/self/verify');
```
rmm5t: Nice cleanup here. 👍

lib/orderbook_sync.js
```diff
@@ -30,20 +30,39 @@ class OrderbookSync extends WebsocketClient {
...
-      this.books[productID] = new Orderbook();
-    });
+    this.productIDs.forEach(this.newProduct, this);
+  }
+
+  newProduct(productID) {
+    this._queues[productID] = [];
+    this._sequences[productID] = [];
+    this.books[productID] = new Orderbook();
   }
```
rmm5t: This method probably deserves to be prefixed with a `_` to designate that it represents more internal/private behavior (i.e. `_newProduct()`).

lib/clients/websocket.js
```diff
@@ -48,28 +48,68 @@ class WebsocketClient extends EventEmitter {
...
+      this.channels.delete(c);
+    }
+
+    this.sendSubscription('unsubscribe', null, channels);
+  }
+
+  onOpen() {
+    this.emit('open');
+
+    this.sendSubscription('subscribe', this.productIDs, this.channels);
```
rmm5t: Perhaps, `sendSubscription()` should default the last two arguments to `this.productIDs` and `this.channels`. i.e. Passing around properties (`this.productIDs` and `this.channels`) as arguments to instance methods that already have access to these properties is bit of a OO smell.

In other words, everytime you want to send a `subscribe` message to re-sync the products and channels, you can just call `this.sendSubscription('subscribe')`.

`sendSubscription()` should probably also be made _"privat"e_ and renamed to `_sendSubscription()`.

lib/orderbook_sync.js
```diff
@@ -30,20 +30,39 @@ class OrderbookSync extends WebsocketClient {
...
+          newProducts = new Set(c.product_ids);
+        }
+      }
+
+      for (let productID of newProducts) {
+        if (productID in this.books) {
+          continue;
+        }
+
+        // add new product
```
rmm5t: superfluous comment

lib/clients/websocket.js
fuzzyTew: I actually had those default arguments, but I took them out because it does create the misleading perception that `sendSubscription('subscribe')` might set the server channels/products to the class ones, when in reality it will only ensure the class ones are subscribed to, but not unsubscribe from any that have been removed.  They must be passed with 'unsubscribe' for that.  But I'll put 'em back!

fuzzyTew:
Thanks for the review.  I'll try to find the time to add all your suggestions and some tests.

fuzzyTew:
I've changed all the mentioned items, written a test for each new API function, and squashed it into one commit.

rmm5t:
Good work @fuzzyTew. I gave it another quick look, and things look much better (especially the additional tests). Thank you. Let's see what @fb55 thinks about these changes.

(I'm good with merging, but still waiting on collaborator/write access to the repository)

rmm5t:
@fuzzyTew Would you also please add a short snippet about these new Websocket usage patterns to the relevant section in the README?
https://github.com/coinbase/gdax-node#websocket-client

rmm5t:
@fb55 Do you think we can get this merged in? I'd like to use the changes here as the basis for making the `WebsocketClient` more resilient by default (auto-reconnects, automatic heartbeat monitoring, etc).

lib/clients/websocket.js
```diff
@@ -48,28 +46,64 @@ class WebsocketClient extends EventEmitter {
...
-        this.channels ? '/users/self/verify' : '/users/self'
-      );
+      let sig = signRequest(this.auth, 'GET', '/users/self/verify');
       Object.assign(subscribeMessage, sig);
     }
 
     this.socket.send(JSON.stringify(subscribeMessage));
   }
 
+  subscribeProduct(...productIDs) {
```
fb55: Not sure I agree with the spirit of this. It should be possible to pick and choose which products to listen to for each channel.

lib/clients/websocket.js
```diff
@@ -17,13 +17,11 @@ class WebsocketClient extends EventEmitter {
...
     super();
-    this.productIDs = Utils.determineProductIDs(productIDs);
+    this.productIDs = new Set(Utils.determineProductIDs(productIDs));
     this.websocketURI = websocketURI;
     this.auth = Utils.checkAuth(auth);
-    this.channels = channels || ['full'];
-    if (!this.channels.includes('heartbeat')) {
-      this.channels.push('heartbeat');
-    }
+    this.channels = new Set(channels || ['full']);
```
fb55: `channels` right now can also be objects with `name` and `product_id` properties, this breaks this.

fuzzyTew:
Thanks @fb55 ... I'm afraid this issue isn't on my radar at the moment anymore, so I might not be available to revamp it to meet the new suggestions, at least not for a while.

fb55:
No problem 😃 – thanks for getting this started, I‘ll try to get this merged next week!

rmm5t:
> `channels` right now can also be objects with `name` and `product_id` properties, this breaks this.

> Not sure I agree with the spirit of this. It should be possible to pick and choose which products to listen to for each channel.

@fb55 brings up some very good points, but I don't think @fuzzyTew's changes break anything that wasn't already _broken_ or _limited_ in the existing `WebsocketClient`.  

```js
    const subscribeMessage = {
      type: 'subscribe',
      product_ids: this.productIDs,
      channels: this.channels,
    };
```

---

However, with all that said, I do see room for meeting @fb55's points, and also further simplifying things. 

Perhaps the method signature should just match the GDAX API more closely: `subscribe()` and `unsubscribe()`. Both just take an _options_ argument that is expected to match [the subscribe API directly](https://docs.gdax.com/#subscribe). So the subscribe signature would look like `subscribe({ product_ids, channels })`.

Proposed Example:

```js
// simple
wsClient.subscribe({ 
  product_ids: ['ETH-USD', 'ETH-EUR'],
  channels: ['level2', 'user']
});

// advanced
wsClient.subscribe({ 
  channels: [
    { name: 'level2', product_ids: ['BTC-USD'] },
    { name: 'ticker', product_ids: ['ETH-BTC', 'LTC-USD'] }
  ]
);

rmm5t:
@fb55, @fuzzyTew  See #213 for a simplified runtime support of `subscribe`/`unsubscribe`

fb55:
Closing in favour of #213

PR #180, opened 2017-12-27
Remove disclaimers from README

rmm5t:
**Extracted from PR #83**

- Disclaimers are already covered by the official MIT LICENSE
- Semantic versioning isn't a feature; it's an expectation of all good software libraries. 

/cc @fb55


fb55:
Awesome, thanks a lot!

PR #181, opened 2017-12-27
Update gitignore and CircleCI node version

rmm5t:
- Adds `yarn.lock` and `.node-version` to `.gitignore` (two common tools that people use, but the config files don't belong in the repository for a shared library)
- Updates `circle.yml` to rely on latest Node v6 LTS version (v6.12.2)

/cc @fb55


fb55:
Nice!

PR #186, opened 2017-12-30
Make WebsocketClient#on signature generic

brunoscopelliti:
error

I've got troubles with the signature of `WebsocketClient#on` method.
Apparently it conflicts with one of its overload. I find in this case more useful the adoption of a generic types.

Any thoughts?

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -253,7 +253,7 @@ declare module 'gdax' {
             auth?: {key:string, secret:string, passphrase:string},
             { channels }?: WebsocketClientOptions );
 
-        on(event: 'message', eventHandler: (data:object) => void);
+        on<T>(event: 'message', eventHandler: (data:T) => void);
         on(event: 'error', eventHandler: (err) => void);
         on(event: 'open', eventHandler: () => void);
         on(event: 'close', eventHandler: () => void);


ConnorChristie:
This shouldn't be an issue. It works just fine for me in VS Code without generics and also I don't think generics are the correct thing to use to fix this issue.

What editor are you using?

brunoscopelliti:
I am using **vscode 1.19.1** on a mac book. The screenshot is vscode's nord theme.
Why do you think generics are not good in this case?
Let's forget for a sec the issue, the signature of `on` is 
```
on(event: 'message', eventHandler: (data:object) => void);
```
In this way `eventHandler`'s parameter is an `object`, and the type system can't know much more about it. On the other hand, a generic signature let the developer suggest the type system a more specific type.
```
websocket.on<TickerFeed>("on", (tickerMessage) => {});
```
Using generics in this case is something I would recommend even in case I haven't had any other issue. I used generics here to take two pigeons with one bean :)

What do you think?

itajaja:
but generics doesn't give you any extra type safety here. What I mean is that this:

```js
websocket.on<TickerFeed>("on", (tickerMessage) => {});
```

is exactly equivalent to:

```js
websocket.on("on", (tickerMessage: TickerFeed) => {});
```

without the need to add a generic

brunoscopelliti:
@itajaja Thanks for the clarification.
I've got your point, and I agree with you that generics are not needed in this case.
Anyway the conflict on the type of `WebsocketClient#on` remains for me. Are you able to reproduce it?

itajaja:
Hi Bruno, sorry but I am not available to reproduce it. Are you sure you are on the latest version of typescript?

brunoscopelliti:
Yes, version 2.6.2, I've just checked.
Also note that the issue is present only when I try to specify the type of `data`:
```js
websocket.on("message", (data: Something) => {
  // ...
});

// Argument of type '"message"' is not assignable to parameter of type '"close"'.
```

But, the following works just fine:
```js
websocket.on("message", (data: object) => {
  // ...
});
```

I guess I could use the following as a workaround for now... but it's not the ideal solution.
```js
websocket.on("message", (data: object) => {
  const foo = data as Something;
  // ...
});
```

itajaja:
I can definitely specify type of data, there is probably something else fishy going on 🤔

fb55:
Thanks a lot @itajaja for looking into this. Happy to revisit this if there is a concrete improvement!

PR #192, opened 2018-01-01
Bugfix/sync server time

sbalhoff:
This synchronizes the time with the server when using the authenticated client. My computer is 70 seconds behind gdax's server, so I was getting `{ message: 'request timestamp expired' }` in the data field every time I called an authenticated api call.

I added two new functions. `startSynchronizingWithServer` calls the time api every `frequencyMilliseconds` to update a time offset variable from the server that can be used when signing requests. Each time this happens, the supplied callback is called to let the user know if synchronization is working. `stopSynchronizingWithServer` turns off this functionality and sets the time offset variable to 0.

I had originally called the time api when an authenticated request was made if `frequencyMilliseconds` had elapsed, but I think this is cleaner.

This pull request does not address the websocket api. I still can't access the authenticated websocket api because of timeouts.

I can update the readme as well to provide information on synchronizing the time with the server's. This is a simplistic example of using the new functions:

```
const Gdax = require('gdax');

const key = 'key';
const secret = 'secret';
const passphrase = 'passphrase';
const apiURI = 'https://api.gdax.com';

const authedClient = new Gdax.AuthenticatedClient(key, secret, passphrase, apiURI);

function startRunning() {
  authedClient.getAccounts().then(data => console.log(data));
}
  
let isSynchronized = false;
const syncTimeMillis = 5 * 60 * 1000;

authedClient.startSynchronizingWithServer(syncTimeMillis, (err, response, data) => {
  if (!isSynchronized && !err && data.success === true) {
    isSynchronized = true;
    startRunning();
  }
});
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -93,6 +93,9 @@ declare module 'gdax' {
             limit: number;
         };
 
+    export type BoolResult = {
+        success: boolean
+    };
 
     export type Account = {
         id: string,
@@ -165,6 +168,9 @@ declare module 'gdax' {
 
     export class AuthenticatedClient {
         constructor(key: string, secret: string, passphrase: string, apiURI: string);
+
+        startSynchronizingWithServer(frequencyMilliseconds: number, callback: callback<BoolResult>);
+        stopSynchronizingWithServer();
 
         getCoinbaseAccounts(callback: callback<CoinbaseAccount[]>)
         getCoinbaseAccounts(): Promise<CoinbaseAccount[]>;
--- a/lib/clients/authenticated.js
+++ b/lib/clients/authenticated.js
@@ -8,6 +8,7 @@ class AuthenticatedClient extends PublicClient {
     this.key = key;
     this.secret = secret;
     this.passphrase = passphrase;
+    this.serverOffsetSeconds = 0;
   }
 
   request(method, uriParts, opts = {}, callback) {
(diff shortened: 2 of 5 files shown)


rmm5t:
>  _My computer is 70 seconds behind gdax's server_

The real question is, why is your client computer's time incorrect? 

> _This [PR] synchronizes the time with the server when using the authenticated client. _

Your time would be better spent configuring your client computer to synchronize your clock with a [time server](https://en.wikipedia.org/wiki/Network_Time_Protocol). Most operating systems make this a checkbox configuration nowadays.

Can you explain why you think a feature like this should be considered vs just correcting the client's time?

---

My vote is to reject and close this PR.

sbalhoff:
> Can you explain why you think a feature like this should be considered vs just correcting the client's time?

1.  A computer's time is used for more than just connecting to gdax. For instance, if two servers required me to be at two different times, I'd have to spoof one of them.
2. It's easier if it works out of the box.

> The real question is, why is your client computer's time incorrect?

Not sure, but it does look like I'm 70 seconds behind UTC, so I should probably fix that anyway.

rmm5t:
> A computer's time is used for more than just connecting to gdax. 

Exactly. That's why your client machine should be set to the **correct** time.

> For instance, if two servers required me to be at two different times, I'd have to spoof one of them.

That doesn't happen unless a server is misconfigured. That would be a _bug_ within that particular server's infrastructure. You don’t go around adding workarounds to other (working) infrastructures.

> It's easier if it works out of the box.

It does work out of the box -- with the expectation that your client machine's operating system was written in the 21st century.

Please close this PR.

Issue #147, opened 2017-11-22
Ability to set timeout for request


tracerstevens:
Please add the ability to set the timeout for the http request.

rmm5t:
If you want a timeout for a specific request, I think you can call one of the HTTP-named methods of the GDAX client directly. For example:

```js
  // Original
  gdax.getProducts();

  // With Timeout
  gdax.get(['products'], { timeout: 1500 }).catch(console.error);
```

PR #198, opened 2018-01-04
Add a default timeout to each request

CjS77:
`request` doesn't have a default timeout set. 
If GDAX API servers time out, the connection will stay open and the callbacks for HTTP requests will never be called leaving requests in limbo.

This PR adds a default timeout of 10s to all requests and adds an option to set the value in the constructor.

### DIFF
--- a/lib/clients/public.js
+++ b/lib/clients/public.js
@@ -1,8 +1,9 @@
 const request = require('request');
 const { Readable } = require('stream');
+const DEFAULT_TIMEOUT = 10 * 1000; // 10 sec
 
 class PublicClient {
-  constructor(apiURI = 'https://api.gdax.com') {
+  constructor(apiURI = 'https://api.gdax.com', options = {}) {
     this.productID = 'BTC-USD';
     if (apiURI && !apiURI.startsWith('http')) {
       process.emitWarning(
@@ -15,6 +16,7 @@ class PublicClient {
 
     this.apiURI = apiURI;
     this.API_LIMIT = 100;
+    this.timeout = +options.timeout > 0 ? options.timeout : DEFAULT_TIMEOUT;
   }
 
   get(...args) {
@@ -86,6 +88,7 @@ class PublicClient {
       method: method.toUpperCase(),
       uri: this.makeAbsoluteURI(this.makeRelativeURI(uriParts)),
       qsStringifyOptions: { arrayFormat: 'repeat' },
+      timeout: this.timeout,
     });
     this.addHeaders(opts);
     const p = new Promise((resolve, reject) => {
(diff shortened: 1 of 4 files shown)


lib/clients/public.js
```diff
@@ -278,9 +281,8 @@ class PublicClient {
     if (!productID || typeof productID !== 'string') {
       process.emitWarning(
         `\`${caller}()\` now requires a product ID as the first argument. ` +
-          `Attempting to use PublicClient#productID (${
-            this.productID
-          }) instead.`,
+          `Attempting to use PublicClient#productID (${this
```
CjS77: A git precommit hook is reformatting this line like this 🤷‍♂️

fb55:
Nice – could you add a test for this?

lib/clients/public.js
rmm5t: Yeah `lint-staged` is obnoxious sometimes.  The weird thing is that the original version was also auto-formatted by `lint-staged` too (it wasn't the original way I wanted this line to look). `¯\_(ツ)_/¯ `

fb55: We should bump prettier, the newer version renders this nicely.

lib/clients/public.js
```diff
@@ -1,8 +1,9 @@
 const request = require('request');
 const { Readable } = require('stream');
+const DEFAULT_TIMEOUT = 10 * 1000; // 10 sec
 
 class PublicClient {
-  constructor(apiURI = 'https://api.gdax.com') {
+  constructor(apiURI = 'https://api.gdax.com', options = {}) {
```
rmm5t: Maybe this is personal preference, but I'd prefer to see the defaults for all options set in the method signature:

```js
constructor(apiURI = 'https://api.gdax.com', { timeout = DEFAULT_TIMEOUT } = {})

rmm5t:
This PR should add an `options` argument to the `AuthenticatedClient` as well.

Also, this PR description needs a reference to #147 with a "Closes #147"

lib/clients/public.js
CjS77: I generally agree, but there's some additional input checking in the constructor in this instance.

CjS77:
Fixes #147

lib/clients/public.js
CjS77: Good catch on the authclient constructor

lib/clients/public.js
CjS77: Done

lib/clients/public.js
rmm5t: > I generally agree, but there's some additional input checking in the constructor in this instance.

 But this input checking is mostly unnecessary and actually invalid. A `timeout` of `0` or `undefined` is perfectly valid using `request` library, and any other validation of `timeout` is already handled by the `request` library. e.g. negative values are already normalized.

PR #199, opened 2018-01-04
Added explicit any types to all untyped code

drupsys:
This was done to silence the ts compiler with strict type checks enabled.

It is just a temporary solution, type definitions should contain few (ideally none) `any` types, hopefully, someone who knows what the correct types are will define them properly and replace this quick fix.


fb55:
Thanks a lot!

PR #205, opened 2018-01-08
Change granularity parameter in README to an acceptable input.

kevinstubbs:
The value '3000' is not an acceptable input for the granularity parameter of getProductHistoricRates. From the [API documentation](https://docs.gdax.com/#get-historic-rates):

> The granularity field must be one of the following values: {60, 300, 900, 3600, 21600, 86400}. Otherwise, your request will be rejected.


fb55:
Thanks a lot!

PR #209, opened 2018-01-08
throw if error response is received

itajaja:
solves https://github.com/coinbase/gdax-node/issues/208

need to add tests, does this approach sound good?

PR #210, opened 2018-01-08
Fix request error handling

### LINKED ISSUE #208: APIs requests don't error out
If an API request returns an error (unable to fulfill order, rate limited, etc), the client will silently return the error payload without erroring out, which makes for an odd exception handling. Is this by design? Sorry if this issue is a duplicate

rmm5t:
## What does it do?

- Fixes error handling expectations as documented in #208
- Non-2xx HTTP status codes will now (correctly) return an error

## Related Issues

- Fixes #208
- Closes #209 (this PR is stacked on top of that one, but this one adds tests and fixes extra edge cases)

### DIFF
--- a/lib/clients/public.js
+++ b/lib/clients/public.js
@@ -58,22 +58,28 @@ class PublicClient {
       } catch (e) {
         data = null;
       }
-      if (typeof callback === 'function') {
-        callback(err, response, data);
-        return;
-      }
+
       if (err) {
         err.response = response;
-        reject(err);
-      }
-      if (data === null) {
+      } else if (response.statusCode > 299) {
+        err = new Error(`HTTP ${response.statusCode} Error`);
+        err.response = response;
+        err.data = data;
+      } else if (data === null) {
         err = new Error('Response could not be parsed as JSON');
         err.response = response;
-        reject(err);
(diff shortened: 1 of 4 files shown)


lib/clients/public.js
```diff
@@ -58,17 +58,28 @@ class PublicClient {
...
-        return;
-      }
+
       if (err) {
         err.response = response;
-        reject(err);
-      }
-      if (data === null) {
+      } else if (response.statusCode > 299) {
+        err = new Error(`HTTP ${response.statusCode} Error`);
```
fb55: Use the `message` from the response here?

rmm5t: Do you mean the `statusMessage` from the `response`, or the `message` from the `body`, or both? I'm inclined to try both.

fb55: `message` from the body. `statusMessage` wouldn't be too bad either 🤷‍♂️

rmm5t:
@fb55 I added the body's error message to the Error object's message. I skipped the `statusMessage`, because it's not all that beneficial anyway and [nock doesn't yet support mocking it at the moment](https://github.com/node-nock/nock/projects/2).

fb55:
Amazing, thanks a lot!

PR #213, opened 2018-01-09
Add Websocket#subscribe() and #unsubscribe()

rmm5t:
## What does it do?

- This PR is stacked on top of @fuzzyTew's PR #179.
- Instead of a stateful `Websocket#[un]subscribeProduct()` and `#[un]subscribeChannel()`, the changes here simplify the method signatures down to just a stateless `#subscribe()` and `#unsubscribe()` that is a direct pass through to the GDAX API.
- These new method signatures match the [GDAX API `subscribe`/`unsubscribe` messages](https://docs.gdax.com/#subscribe) directly.

## Related Issues

- Closes #179 

## Example Usage:

```javascript
websocket.unsubscribe({ channels: ['full'] });

websocket.subscribe({ product_ids: ['LTC-USD'], channels: ['ticker', 'user'] });

websocket.subscribe({
  channels: [{
    name: 'user',
    product_ids: ['ETH-USD']
  }]
});

websocket.unsubscribe({
  channels: [{
    name: 'user',
    product_ids: ['LTC-USD']
  }, {
    name: 'user',
    product_ids: ['ETH-USD']
  }]
});
```

### DIFF
--- a/lib/clients/websocket.js
+++ b/lib/clients/websocket.js
@@ -17,11 +17,13 @@ class WebsocketClient extends EventEmitter {
     { channels = null } = {}
   ) {
     super();
-    this.productIDs = new Set(Utils.determineProductIDs(productIDs));
+    this.productIDs = Utils.determineProductIDs(productIDs);
     this.websocketURI = websocketURI;
     this.auth = Utils.checkAuth(auth);
-    this.channels = new Set(channels || ['full']);
-    this.channels.add('heartbeat');
+    this.channels = channels || ['full'];
+    if (!this.channels.includes('heartbeat')) {
+      this.channels.push('heartbeat');
+    }
     this.connect();
   }
 
@@ -50,11 +52,11 @@ class WebsocketClient extends EventEmitter {
     const message = { type };
 
     if (channels) {
-      message.channels = [...channels];
+      message.channels = channels;
     }
 
     if (product_ids) {
-      message.product_ids = [...product_ids];
+      message.product_ids = product_ids;
     }
 
     // Add Signature
--- a/lib/clients/websocket.js
+++ b/lib/clients/websocket.js
@@ -46,62 +46,37 @@ class WebsocketClient extends EventEmitter {
     this.socket = null;
   }
 
-  _sendSubscription(
-    type,
-    productIDs = this.productIDs,
-    channels = this.channels
-  ) {
-    const subscribeMessage = {
-      type: type,
-      channels: [...channels],
-    };
+  _sendSubscription(type, { product_ids, channels }) {
+    const message = { type };
 
-    if (productIDs) {
-      subscribeMessage.product_ids = [...productIDs];
+    if (channels) {
+      message.channels = [...channels];
+    }
+
+    if (product_ids) {
+      message.product_ids = [...product_ids];
     }
 
     // Add Signature
     if (this.auth.secret) {
-      let sig = signRequest(this.auth, 'GET', '/users/self/verify');
-      Object.assign(subscribeMessage, sig);
+      const sig = signRequest(this.auth, 'GET', '/users/self/verify');
+      Object.assign(message, sig);
     }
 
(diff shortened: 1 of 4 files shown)


rmm5t:
Just rebased this against all the new changes in `master`.

lib/clients/websocket.js
```diff
@@ -48,26 +46,37 @@ class WebsocketClient extends EventEmitter {
...
+    this._sendSubscription('subscribe', { product_ids, channels });
+  }
+
+  unsubscribe({ product_ids, channels }) {
+    this._sendSubscription('unsubscribe', { product_ids, channels });
+  }
+
+  onOpen() {
+    this.emit('open');
+    this.subscribe({ product_ids: this.productIDs, channels: this.channels });
```
fb55: `productIDs` and `channels` are both sets

lib/clients/websocket.js
```diff
@@ -48,26 +46,37 @@ class WebsocketClient extends EventEmitter {
     this.socket = null;
   }
 
-  onOpen() {
-    this.emit('open');
+  _sendSubscription(type, { product_ids, channels }) {
+    const message = { type };
+
+    if (channels) {
+      message.channels = [...channels];
```
fb55: Can we do this in `onOpen`? No real need here to create copies of them.

tests/websocket.spec.js
```diff
@@ -99,6 +99,117 @@ suite('WebsocketClient', () => {
     });
   });
 
+  test('subscribes to additional products', done => {
+    var client;
+    const server = testserver(++port, () => {
```
fb55: Didn't we stop incrementing the port?

tests/websocket.spec.js
```diff
@@ -99,6 +99,117 @@ suite('WebsocketClient', () => {
     });
   });
 
+  test('subscribes to additional products', done => {
+    var client;
```
fb55: `let`

lib/clients/websocket.js
```diff
@@ -17,13 +17,11 @@ class WebsocketClient extends EventEmitter {
     { channels = null } = {}
   ) {
     super();
-    this.productIDs = Utils.determineProductIDs(productIDs);
+    this.productIDs = new Set(Utils.determineProductIDs(productIDs));
```
fb55: Not sure what these sets are used for — why aren't we using plain objects?

lib/clients/websocket.js
rmm5t: I think the copies were because Sets were being used. I'll kill 'em.

tests/websocket.spec.js
rmm5t: True. This must have been left over from multiple rebases. I'll fix.

tests/websocket.spec.js
rmm5t: Good catch. I'll fix.

lib/clients/websocket.js
rmm5t: Sets are a better construct for unique lists (obviously), but if you don't like them, I'll clean them out.

rmm5t:
@fb55 Requested changes applied.

lib/orderbook_sync.js
```diff
@@ -67,8 +69,21 @@ class OrderbookSync extends WebsocketClient {
       .catch(problems);
   }
 
+  // subscriptions changed -- possible new products
+  _newSubscription(data) {
+    const channel = data.channels.find(c => c.name === 'full');
+    channel && channel.product_ids
+      .filter(productID => !(productID in this.books))
+      .forEach(productID => this._newProduct(productID));
```
fb55: Use same pattern as above (`this.productIDs.forEach(this._newProduct, this);`)?

rmm5t: I dig it. Good eye. Done.

fb55:
Awesome, thanks :)

PR #217, opened 2018-01-10
Add a function to fetch transfers in the account

adityamertia:
This function can be used to fetch transfers done in an account. Useful in fetching txHash of a transfer made.

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,6 +260,16 @@ authedClient.getAccountHistory(accountID, callback);
 authedClient.getAccountHistory(accountID, { before: 3000 }, callback);
 ```
 
+* [`getAccountTransfers`](https://docs.gdax.com/#get-account-transfers)
+
+```js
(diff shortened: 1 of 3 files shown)


rmm5t:
@adityamertia Please add tests.

adityamertia:
@rmm5t Kindly review.

fb55:
This is an interesting case, as it is something the GDAX API simply proxies from Coinbase's API. Ideally you'd query Coinbase directly, eg. through [`coinbase/coinbase-node`](https://github.com/coinbase/coinbase-node). The user experience for this isn't great though, so there is a point about documenting our internal endpoints.

I'll add this to the agenda for our next internal team sync, let's see what people think.

fb55:
Sorry for the delay. These endpoints are going to be documented, so this is good to go! :D

PR #221, opened 2018-01-11
Emit error when websocket message is an error

### LINKED ISSUE #219: `message` errors not passed to the `error` handler
Probably related to #208. For example, subscribing with an invalid product_id returns an error to the raw on('message') socket, but not to the library.

```js
const gdax = require('gdax');
const websocket = new gdax.WebsocketClient(['ETH-GBP']);

websocket.on('message', data => { console.log(data.product_id, data.price); });  // undefined undefined
websocket.on('error', err => { console.error(err); });  // not reached
websocket.on('close', () => { console.error('close'); });
```

That sort of error (invalid product_id) is an application level one, which is why I guess it's returned in the `message` payload. However, to the user, this error is as critical as a websocket-level connectivity error, and handling it in `error` would be easiest.

rmm5t:
## What does it do?

Before, if an `error` message is sent from the server, that error is emitted as regular `message` data. Now, if a message of `type=error` arrives, it is emitted as an `error` instead.

## What else?

Also removes auto-incrementing port for testing websocket server connections. We got to a point where we ran into port conflicts on the CI server.

## Related Issues

- FIxes #219

/cc @fb55

### DIFF
--- a/lib/clients/websocket.js
+++ b/lib/clients/websocket.js
@@ -76,7 +76,12 @@ class WebsocketClient extends EventEmitter {
   }
 
   onMessage(data) {
-    this.emit('message', JSON.parse(data));
+    const message = JSON.parse(data);
+    if (message.type === 'error') {
+      this.onError(message);
+    } else {
+      this.emit('message', message);
+    }
   }
 
   onError(err) {
--- a/tests/orderbook_sync.spec.js
+++ b/tests/orderbook_sync.spec.js
@@ -10,7 +10,7 @@ const EXCHANGE_API_URL = 'https://api.gdax.com';
 
 suite('OrderbookSync', () => {
   test('not passes authentication details to websocket', done => {
-    const server = testserver(++port, () => {
+    const server = testserver(port, () => {
       new Gdax.OrderbookSync(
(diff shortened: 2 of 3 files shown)


fb55:
How do you feel about emitting an event with the type as the name as well? So instead of creating a special case for error just also create matches, open, l2update etc. events.

rmm5t:
> How do you feel about emitting an event with the type as the name as well? So instead of creating a special case for error just also create matches, open, l2update etc. events.

🤔 My initial gut reaction was to say no, but after considering how simple the implementation would be (simpler than this PR's changes), and as long as the generic `message` event is _also_ emitted at the same time, I don’t really have any strong arguments that I can articulate at the moment. 

Long story short, as long as `message` events remain for backwards compatibility and generic behavior, I think specific event types could be a nicety.

I'll modify this PR to be more generic like that.

rmm5t:
@fb55 Oh. Here's an argument against the specfic event types. What about the websocket's `open` events (for Orders). This would conflict with the socket 'open' events.

fb55:
Whee, good point. Merging as is, thanks a lot :)

Issue #226, opened 2018-01-14
WebsocketClient.disconnect() issue


macbeth8c:
Hello,
If one calls WebsocketClient.disconnect() method (which is not officially exposed in the interface, I know :), the disconnect or onClose methods invalidate the socket reference before other functions wish to use it, e.g. remaining data receive (the disconnect does not occur immediately). The error is thrown:

> Could not disconnect (not connected)
Or null pointer exceptions stating that socket is null already.

. I fixed it by commenting the following lines & added a different conditions.
```
  connect() {
    if (this.socket && this.socket.readyState === Websocket.OPEN) {
      this.socket.close();
    }

    this.socket = new Websocket(this.websocketURI);

    this.socket.on('message', this.onMessage.bind(this));
    this.socket.on('open', this.onOpen.bind(this));
    this.socket.on('close', this.onClose.bind(this));
    this.socket.on('error', this.onError.bind(this));
  }

  disconnect() {
    if (!this.socket || this.socket.readyState !== Websocket.OPEN) {
      throw new Error('Could not disconnect (not connected)');
    }
    this.socket.close();
    //this.socket = null;
  }
//...
  onClose() {
    //this.socket = null;
    this.emit('close');
  }
```
In my snippets you can see another way to check for connection status using ws.OPEN.

Cheers

rmm5t:
> the disconnect or onClose methods invalidate the socket reference before other functions wish to use it, e.g. remaining data receive (the disconnect does not occur immediately).

Could you elaborate on this more? I cannot reproduce your issue. What other methods are trying to use the socket reference after `close` is emitted? For example, the only methods that reference `this.socket` while a socket is potentially open are `onClose()` and `disconnect()`.

For example, to test, I set an interval timer to call `disconnect()` every second while reading from the `full` channel, and I cannot trigger the same behavior that you're seeing. Btw, I run my websocket under a "Feed" wrapper that automatically reconnects whenever it recognizes that the underlying websocket was disconnected (via `close` event).

macbeth8c:
Hi,
to reproduce the issue in the easiest way I use the following code
```
websocket.on('message', processTickerData);
websocket.on('error', err => { console.log(util.inspect(err, { showHidden: false, depth: null })) });
websocket.on('close', () => { console.log('Stream closed.') });
websocket.on('open', () => { console.log('Stream opened.') });


setInterval(function () {
  console.log('reopen attempt.')
  websocket.connect();
}, 2000);
```
```
[...]

PR #229, opened 2018-01-16
Fix connect() and disconnect() behaviour (#226)

macbeth8c:
Fix connect() and disconnect() behaviour (#226)

### DIFF
--- a/lib/clients/websocket.js
+++ b/lib/clients/websocket.js
@@ -28,8 +28,8 @@ class WebsocketClient extends EventEmitter {
   }
 
   connect() {
-    if (this.socket) {
-      this.socket.close();
+    if (this.socket && this.socket.readyState !== Websocket.CLOSED) {
+      throw new Error('Could not connect (not disconnected)');
     }
 
     this.socket = new Websocket(this.websocketURI);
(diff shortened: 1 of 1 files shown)


lib/clients/websocket.js
```diff
@@ -41,11 +41,10 @@ class WebsocketClient extends EventEmitter {
   }
 
   disconnect() {
-    if (!this.socket) {
+    if (!this.socket || this.socket.readyState !== Websocket.OPEN) {
```
fb55: Should also consider `CONNECTING`

lib/clients/websocket.js
```diff
@@ -28,8 +28,8 @@ class WebsocketClient extends EventEmitter {
   }
 
   connect() {
-    if (this.socket) {
-      this.socket.close();
+    if (this.socket && this.socket.readyState !== Websocket.CLOSED) {
```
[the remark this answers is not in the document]
macbeth8c: The CLOSING was omitted on purpose. What is a meaning of connecting when the precious connection is not yet closed? Please explain

lib/clients/websocket.js
macbeth8c: The CONNECTING was omitted on purpose. The meaning here might be that someone wants to abort the connection. In this case the socket will abort and throw an exception 'WebSocket was closed before the connection was established'. If this is the behaviour we want, I can fix it.

lib/clients/websocket.js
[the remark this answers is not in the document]
fb55: The socket isn't usable anymore if the state is `CLOSING`, so calling `connect` should be legitimate.

lib/clients/websocket.js
fb55: Interesting. That error might still be more descriptive than the current message, but I'm fine with omitting it.

PR #230, opened 2018-01-18
Fix #225

### LINKED ISSUE #225: WebsocketClient interface connect/disconnect
Imho it would be useful to have websocket client's methods `connect` and `disconnect` exposed on the WebsocketClient interface.
What do you think?

brunoscopelliti:
Exposed `connect` and `disconnect` on websocket client interface.


fb55:
Nice, thanks!

PR #234, opened 2018-01-19
Ignore some files when publishing the package

EddyVerbruggen:
No need to include the CI config and `tests` folder in the published npm package 😘

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,2 +1,9 @@
+*.swp
+
+/.idea
+/package-lock.json
+/yarn.lock
+/.node-version
+
 circle.yml
-tests/
+/tests


fb55:
`.npmignore` files override `.gitignore` files (see https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package), so most of the contents of the `.gitignore` (with the exception of `node_modules`) also need to be copied to it.

In general I always feel doing this results in very little gains and is probably not worth the overhead that is remembering to add rules to both files.

rmm5t:
> .npmignore files override .gitignore files

Good point. I always forget this too, until it bites me.

EddyVerbruggen:
Thanks, @fb55, that's a very good point. I usually have a folder structure where the npm distributable lives in a separate folder so I don't need to think about what contents of `.gitignore` get overridden, but that's of course not applicable to this repo.

Btw, I ran `npm pack` to test the effect of this change to the package filesize (zipped):
- Without `.npmignore`: 63 KB
- With `.npmignore`: 18 KB

@rmm5t Can you please use the 'Squash and merge' option so I don't need to bother squashing these commits? Thank you!

screen shot 2018-01-19 at 22 04 35

fb55:
@EddyVerbruggen Imho 45KB aren't enough of a reason to force all contributors to remember this. We can revisit this once the size actually impacts download speeds.

EddyVerbruggen:
OK, no worries. 

It was never about the size really, I just like my packages as clean as possible. Mine are usually shipped with webapps and mobile apps so I have to always focus on this (smallest possible package) aspect (otherwise users will complain). 

Would be cool if somehow .npmignore could extend .gitignore - then this would be a no-brainer.

Keep up the good work folks!

rmm5t:
> Would be cool if somehow .npmignore could extend .gitignore - then this would be a no-brainer.

Right?!?!  I've wished NPM and Yarn would get their act together on this topic. [Yarn has other `.npmignore` issues](https://github.com/yarnpkg/yarn/issues/2090).

PR #237, opened 2018-01-19
Improve error handling by checking auth before every signature

### LINKED ISSUE #232: Error Handling when Auth Data undefined
I ran in to some errors when first writing my code, and I think the AuthClient needs handle Authentication a little better.

When one of the varibles are "undefined" the gdax api response with a Buffer error 
So for example this code ( authClient object printed out for visibility)

> var Gdax = require('gdax');
> var gdaxconfig = require('./gdax.config')
> const key = gdaxconfig.key;
> const secret = gdaxconfig.secret;
> const passphrase = gdaxconfig.passphrase;
> 
> const liveapiURI = 'https://api.gdax.com';
> const sandboxURI = 'https://api-public.sandbox.gdax.com';
> var apiURI=sandboxURI;
> const authedClient = new Gdax.AuthenticatedClient(key, secret, passphrase, apiURI);
> console.log(authedClient);
> authedClient.getAccounts((error,response,data)=>{
> });

As you can see in this example the b64secret is undefined

> AuthenticatedClient {
>   productID: 'BTC-USD',
>   apiURI: 'https://api-public.sandbox.gdax.com',
>   key: 'XXXXXXXXXXXXXXXXXXXXX',
>   b64secret: undefined,
>   passphrase: 'YYYYYYY' }
> 

It produces this Exception

> buffer.js:183
>     throw new TypeError(kFromErrorMsg);
>     ^
> 
> TypeError: First argument must be a string, Buffer, ArrayBuffer, Array, or array-like object.
>     at Function.Buffer.from (buffer.js:183:11)
>     at Buffer (buffer.js:158:17)
>     at module.exports.signRequest (/home/eperry/gdax/swinbot/node_modules/gdax/lib/request_signer.js:24:17)
>     at AuthenticatedClient.prototype._getSignature (/home/eperry/gdax/swinbot/node_modules/gdax/lib/clients/authenticated.js:60:15)
>     at AuthenticatedClient.prototype.request (/home/eperry/gdax/swinbot/node_modules/gdax/lib/clients/authenticated.js:50:32)
>     at apply (/home/eperry/gdax/swinbot/node_modules/lodash._createwrapper/index.js:72:25)
>     at AuthenticatedClient.wrapper [as get] (/home/eperry/gdax/swinbot/node_modules/lodash._createwrapper/index.js:415:12)
>     at AuthenticatedClient.prototype.getAccounts (/home/eperry/gdax/swinbot/node_modules/gdax/lib/clients/authenticated.js:78:26)
>     at Object.<anonymous> (/home/eperry/gdax/swinbot/test-auth.js:15:14)
>     at Module._compile (module.js:635:30)
> 

This is a very confusing exception, where I think the values of key, secret and passphrase should be tested at minimum to be something other then undefined. Maybe something like

>if ( key === undefined && typeof key === "string" ){
>    // Do some error message
>}

rmm5t:
Results in the following error if credentials are missing before signing a request:

```
Error: Invalid or incomplete authentication credentials. Please provide all of the key, secret, and passphrase fields.
```

Closes #232

/cc @fb55

### DIFF
--- a/lib/request_signer.js
+++ b/lib/request_signer.js
@@ -1,6 +1,8 @@
 'use strict';
 const crypto = require('crypto');
 const querystring = require('querystring');
+const Utils = require('./utilities.js');
+
 /**
  Signs request messages for authenticated requests to GDAX
  * @param auth {object} hash containing key, secret and passphrase
@@ -12,6 +14,7 @@ const querystring = require('querystring');
  * @returns {{key: string, signature: *, timestamp: number, passphrase: string}}
  */
 module.exports.signRequest = (auth, method, path, options = {}) => {
+  Utils.checkAuth(auth);
   const timestamp = Date.now() / 1000;
   let body = '';
   if (options.body) {
--- a/lib/utilities.js
+++ b/lib/utilities.js
@@ -13,9 +13,9 @@ function determineProductIDs(productIDs) {
 }
 
 function checkAuth(auth) {
-  if (auth && !(auth.secret && auth.key && auth.passphrase)) {
+  if (auth && !(auth.key && auth.secret && auth.passphrase)) {
     throw new Error(
-      'Invalid or incomplete authentication credentials. You should either provide all of the secret, key and passphrase fields, or leave auth null'
+      'Invalid or incomplete authentication credentials. Please provide all of the key, secret, and passphrase fields.'
     );
   }
   return auth || {};


edperry:
IMO this is better because now we are checking that key is actually set to something

As much as I understand the coding,  I thought I would see something like
> if ( typeof auth.key !== 'undefined' && auth.key )
but you probably understand nodejs/code base better.  

Thanks for the fast turn around

rmm5t:
> `if ( typeof auth.key !== 'undefined' && auth.key )`

^^^ Mostly a waste of keystrokes and unnecessary brain garbling to understand the intent. Not to mention that it fails to check if `auth` is even defined.

> `if (auth && auth.key)`

^^^ Readable, accomplishes everything you need, and has the side benefit of also checking against `null` and empty strings.

fb55:
Awesome, thanks!

PR #240, opened 2018-01-19
Fix precommit: eslint cannot be run on json or markdown

rmm5t:
Fixes mistake in `precommit` setup. `eslint` cannot be run on json or markdown.

Ref #235

/cc @fb55

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,10 +65,7 @@
     "precommit": "lint-staged"
   },
   "lint-staged": {
-    "*.{js,json,md}": [
-      "eslint --fix",
-      "prettier --write",
-      "git add"
-    ]
+    "*.{js}": ["eslint --fix", "prettier --write", "git add"],
+    "*.{json,md}": ["prettier --write", "git add"]
   }
 }


rmm5t:
@fb55 I was trying to upgrade core dependencies when I ran into this issue. 

**Note**: This _must_ be merged before any further changes to `package.json` can be made, because it effectively locks out all `*.json` and `*.md` commits (unless you disable the precommit hook)

package.json
```diff
@@ -65,10 +65,7 @@
     "precommit": "lint-staged"
   },
   "lint-staged": {
-    "*.{js,json,md}": [
-      "eslint --fix",
-      "prettier --write",
-      "git add"
-    ]
+    "*.{js}": ["eslint --fix", "git add"],
```
fb55: Change this to `"*.js": ["eslint --fix", "prettier --write", "git add"],` and remove `js` from below?

fb55:
Whee, bad mistake — thanks for flagging.

package.json
rmm5t: No problem. Done.

PR #239, opened 2018-01-19
Fix UnhandledPromiseRejectionWarning in OrderbookSync

rmm5t:
I got tired of seeing the following warning in the test suite:

```
(node:59176) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 3): Error: Failed to load orderbook: whoops
```

Turned out that simplifying the code behind grabbing the initial Orderbook did the trick.

I think this fixes #233 

/cc @fb55

PR #241, opened 2018-01-20
Upgrade and relax dependencies

rmm5t:
## What does it do?

- Upgrades and relaxes core dependencies. Static versions are ill-advised for shared libraries, because clients cannot take advantage of 3rd party bug fixes and performance improvements. Let semver do its thing.
- Upgrades mocha and nock dev dependencies. This actually exposed an errant test connection in test suite.
- Closes out the errant websocket server in test suite.

/cc @fb55 

Aside, much of ~#239~ (now merged) and ~#240~ (now merged) was discovered while trying to work on this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,8 +31,8 @@
     "eslint": "^4.0.0",
     "husky": "^0.13.4",
     "lint-staged": "^3.6.1",
-    "mocha": "3.4.2",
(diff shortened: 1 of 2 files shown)


fb55:
Nice, thanks a lot :)

PR #243, opened 2018-01-20
Add clockSkew parameter to account for skew between GDAX time and local machine time

prabhakar97:
## Scenario
If a user's local machine's time is skewed with respect to GDAX API, the API calls fail with error response `{ message: 'request timestamp expired' }`. A solution is for the user to adjust their machine's time. But if a user is unwilling to do that, s/he should have a way to configure the authenticated client to account for the clock skew.

## Change summary
This commit adds support for a `clockSkew` field in options parameter of the AuthenticatedClient constructor whose value should be set to the difference of time between GDAX API server (can be found as epoch field at https://api.gdax.com/time) and local machine's time (found using Date.now() in Node or browser).

### DIFF
--- a/lib/clients/authenticated.js
+++ b/lib/clients/authenticated.js
@@ -8,12 +8,18 @@ class AuthenticatedClient extends PublicClient {
     this.key = key;
     this.secret = secret;
     this.passphrase = passphrase;
+    if (options.offset > 0) {
+      this.offset = options.offset;
+    }
   }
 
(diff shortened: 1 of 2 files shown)


fb55:
Having a somewhat accurate clock is something a lot of applications depend on. Please fix this on the OS level (for example by specifying an NTP server), I wouldn't consider this in the scope of this library.

PR #245, opened 2018-01-23
Faster decimal library.

lisbakke:
I was loading a lot of a data and noticed that the `num` library was slow. I gave `decimal.js` a shot and it's significantly faster.

### DIFF
--- a/lib/orderbook.js
+++ b/lib/orderbook.js
@@ -1,5 +1,5 @@
 const { RBTree } = require('bintrees');
-const num = require('num');
+const BigNumber = require('bignumber.js');
 const assert = require('assert');
 
 class Orderbook {
@@ -19,8 +19,8 @@ class Orderbook {
         .map(order => ({
           id: order[2],
           side: 'buy',
-          price: num(order[0]),
-          size: num(order[1]),
+          price: BigNumber(order[0]),
+          size: BigNumber(order[1]),
         }))
         .forEach(order => this.add(order));
 
@@ -28,8 +28,8 @@ class Orderbook {
         .map(order => ({
           id: order[2],
           side: 'sell',
-          price: num(order[0]),
-          size: num(order[1]),
+          price: BigNumber(order[0]),
+          size: BigNumber(order[1]),
         }))
         .forEach(order => this.add(order));
     } else {
@@ -50,8 +50,8 @@ class Orderbook {
     order = {
       id: order.order_id || order.id,
       side: order.side,
-      price: num(order.price),
-      size: num(order.size || order.remaining_size),
+      price: BigNumber(order.price),
+      size: BigNumber(order.size || order.remaining_size),
     };
 
     const tree = this._getTree(order.side);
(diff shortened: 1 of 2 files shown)


rmm5t:
> I gave decimal.js a shot and it's significantly faster.

I'm in favor of dropping num (because it's a risk from the perspective of canonical popularity), but do you have any benchmark metrics to prove this claim? 

Personally, I'd love to see a benchmark comparison between [num](https://github.com/defunctzombie/num), [decimal.js](https://github.com/MikeMcl/decimal.js), [big.js](https://github.com/MikeMcl/big.js/), and [bignumber.js](https://github.com/MikeMcl/bignumber.js/).

If nothing else, big.js and bignumber.js win from a familiarity and popularity perspective (according to NPM downloads and github stars).

rmm5t:
This got me searching. Interestingly, while _decimal.js_ was written to be "v2" of _bigdecimal.js_, the author of both claims that _bigdecimal.js_ is potentially better for financial applications; whereas, _decimal.js_ is potentially better for scientific applications:

https://github.com/MikeMcl/big.js/issues/45#issuecomment-104211175

> _decimal.js_ was orginally intended to be version two of _bignumber.js_ but I decided to make them separate libraries. The main difference between them is that with _decimal.js_ precision is specified in terms of significant digits instead of decimal places, and all calculations are rounded to the precision (similar to Python's decimal module) rather than just those involving division.

> This could be considered to make _bignumber.js_ easier to use than _decimal.js_, and perhaps more suitable for financial applications, because the user doesn't need to worry about losing precision unless an operation involving division is used.

> Conversely, _decimal.js_ may be considered better for more scientifc applications as it can handle very small or large values more efficiently: for example, it does not have the limitation of _bignumber.js_ that if adding a BigNumber value with a small exponent to one with a large exponent, _bignumber.js_ will attempt to perform the operation to full precision which may involve the creation of a massive array, and consequently cause the operation to hang.

lisbakke:
I could provide metrics for `decimal.js` vs. `num` in the morning if you'd like. It's near 10x faster when loading the level 3 order book for the first time at https://github.com/coinbase/gdax-node/blob/master/lib/orderbook.js#L18

lisbakke:
I just did 2 runs to time `orderbook.js` line 18 -> 34 (loading new orderbook).

`decimal.js` 850ms & 805ms
`num` 12,200ms & 12,300ms

rmm5t:
> It's near 10x faster when loading the level 3 order book for the first time

Very nice! However, given that _decimal.js_ only handles precision rounding (significant digits), but _bignumber.js_ handles decimal place rounding, I think  _bignumber.js_ is a much better fit for a cryptocurrency/financial library.

Thoughts?

lisbakke:
Changed to `BigNumber.js`, performance looks good :+1:

nikulis:
BigNumber is also what [GDAX Trading Toolkit](https://github.com/coinbase/gdax-tt) uses, so that consistency between Coinbase projects would be nice. (my 2¢)

fb55:
This looks great. I agree that we should probably switch to another library — BigNumber or Big.js would both be two great options.

fb55:
Missed the update — great addition, thanks a lot!

Issue #231, opened 2018-01-18
Types inconsistencies


brunoscopelliti:
At the moment *.d.ts has some inconsistencies.
I am happy to contribute to improve it.

Another example, that is causing trouble to me, and possibly others:

```
getFills(pageArgs: PageArgs, callback: callback<any>): void;
getFills(pageArgs: PageArgs): Promise<any>; 
```

According to [docs](https://docs.gdax.com/#list-fills) I can pass an object with a `product_id` property to `getFills` to get the fills of a particular product (eg. BTC-EUR)...
It works at runtime, but it fails to pass typescript build. 

There's the same problem on [getOrders](https://docs.gdax.com/#list-orders), too.

Or again, the definition of `Account`:

```
  export type Account = {
      id: string,
      profile_id: string,
      currency: CurrencyType,
      balance: number,
      available: number,
      hold: number
  };
```
`balance`, `available` and `hold` are strings.

fb55:
Yup, the interface wasn’t updated to allow this

brunoscopelliti:
If you are ok, I can submit another PR in the next days...

fb55:
@brunoscopelliti That would be dope

brunoscopelliti:
@fb55 do you know when a new version will be published on npm?

PR #248, opened 2018-01-24
Fix types inconsistencies

brunoscopelliti:
Fix type Account

Extend getOrders/getFills signature to accept props beyond pageargs

(see #231)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -78,29 +78,27 @@ declare module 'gdax' {
     }
 
     export type PageArgs = {
-        before: number;
+        before?: number;
         after?: number;
         limit?: number;
-    } |
-        {
-            before?: number;
-            after: number;
-            limit?: number;
-        } |
-        {
-            before?: number;
-            after?: number;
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -78,29 +78,27 @@ declare module 'gdax' {
     }
 
     export type PageArgs = {
```
fb55: One of these is required — shouldn't that be enforced?

brunoscopelliti: Currently, I am using `getOrders` and `getFills` without specifying any of those parameter, and the result I get back is reasonably correct. I am not sure those parameters are mandatory; am I wrong?

fb55: Right, they are only required if you want to access another page :D

fb55:
Awesome, thanks a lot!

brunoscopelliti:
Hey @fb55 ,
Do you have any plan about publishing a new package on npm, containing this fix?

PR #253, opened 2018-01-30
More efficient orderbook sync, emitting 'synced' message upon completion

ko0f:


fb55:
Nice! Would you mind adding a test case for the `synced` message? Also, if we emit this every time we sync we should probably also add an event for when we decide to sync again.

ko0f:
There ya go. :)

fb55:
Amazing, thanks a lot!

PR #266, opened 2018-02-13
corrected README.md Websocket Client example to use wss endpoint

dereksan:
Websocket Client example call should use a wss endpoint, not the rest endpoint.


fb55:
Whee, good catch. Thanks a lot!

PR #268, opened 2018-02-18
Check for null price on change request

### LINKED ISSUE #255: Error thrown from OrderbookSync receiving `change` with undefined price
Using master, it seem that under some condition, OrderbookSync will attempt to process a `change` message that contains no price (no idea why GDAX would be sending such a message):
```
\node_modules\bignumber.js\bignumber.js:1223
            throw error;
            ^

BigNumber Error: new BigNumber() not a number: undefined
    at raise (\node_modules\bignumber.js\bignumber.js:1191:25)
    at \node_modules\bignumber.js\bignumber.js:1179:33
    at new BigNumber \node_modules\bignumber.js\bignumber.js:194:67)
    at BigNumber \node_modules\bignumber.js\bignumber.js:164:24)
    at Orderbook.change \node_modules\gdax\lib\orderbook.js:116:19)
    at OrderbookSync.processMessage (\node_modules\gdax\lib\orderbook_sync.js:132:14)
    at emitOne (events.js:121:20)
    at OrderbookSync.emit (events.js:211:7)
    at OrderbookSync.onMessage (\node_modules\gdax\lib\clients\websocket.js:94:12)
    at emitOne (events.js:116:13)
    at WebSocket.emit (events.js:211:7)
    at Receiver._receiver.onmessage \node_modules\ws\lib\websocket.js:145:47)
    at Receiver.dataMessage (\node_modules\ws\lib\receiver.js:430:14)
    at Receiver.getData (\node_modules\ws\lib\receiver.js:368:12)
    at Receiver.startLoop (\node_modules\ws\lib\receiver.js:164:16)
    at Receiver.add (\node_modules\ws\lib\receiver.js:138:10)
    at TLSSocket._ultron.on (\node_modules\ws\lib\websocket.js:143:54)
    at emitOne (events.js:116:13)
    at TLSSocket.emit (events.js:211:7)
    at addChunk (_stream_readable.js:263:12)
    at readableAddChunk (_stream_readable.js:250:11)
    at TLSSocket.Readable.push (_stream_readable.js:208:10)
```
I haven't yet put in debugging code to try and catch the offending `change` message, but since BigNumber will actually throw an error, perhaps it would be good to wrap some of the calls to it and catch the error and emit an error event from the OrderbookSync instead of throwing the error up the stack. I've had this happen a few times in the past 24 hours, crashing out the application it was running within.

kyle-mccarthy:
Fixes #255.  A null price can be returned from the API on a change request, when this happens in indicates that the order was a market order.  The problem currently is that there is no check to ensure that the price is not null, and passing a null price to BigNumber will cause an uncaught exception.

> Any change message where the price is null indicates that the change message is for a market order. Change messages for limit orders will always have a price specified.

[Relevant GDAX API docs](https://docs.gdax.com/#the-code-classprettyprintfullcode-channel)

### DIFF
--- a/lib/orderbook.js
+++ b/lib/orderbook.js
@@ -17,18 +17,18 @@ class Orderbook {
     if (book) {
       book.bids
         .forEach(order => this.add({
-           id: order[2],
-           side: 'buy',
-           price: BigNumber(order[0]),
-           size: BigNumber(order[1]),
+          id: order[2],
+          side: 'buy',
+          price: BigNumber(order[0]),
+          size: BigNumber(order[1]),
         }));
 
       book.asks
         .forEach(order => this.add({
-           id: order[2],
-           side: 'sell',
-           price: BigNumber(order[0]),
-           size: BigNumber(order[1]),
+          id: order[2],
+          side: 'sell',
+          price: BigNumber(order[0]),
+          size: BigNumber(order[1]),
         }));
     } else {
       book = { asks: [], bids: [] };
@@ -110,6 +110,11 @@ class Orderbook {
   }
 
   change(change) {
+    // price of null indicates market order
+    if (change.price === null || change.price === undefined) {
+      return;
+    }
+
     const size = BigNumber(change.new_size);
     const price = BigNumber(change.price);
     const order = this.get(change.order_id);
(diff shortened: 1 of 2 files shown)


kyle-mccarthy:
@rmm5t Okay no problem, I added a very basic test case for this scenario.

fb55:
Thanks a lot!

PR #282, opened 2018-03-21
Arg count fix in readme

wbt:
"Your callback should accept **two** arguments:" but then three are listed.  This addresses the inconsistency.


fb55:
Good catch, thanks!

PR #284, opened 2018-03-22
Add typings for WebsocketClient, AuthenticatedClient

eddiew:


index.d.ts
```diff
@@ -251,6 +254,66 @@ declare module 'gdax' {
...
+        export type L2Snapshot = {
+            type: 'snapshot'
+            product_id: string
+            bids: [string, string][] // strings are serialized fixed-point numbers
+            asks: [string, string][]
+        }
+        export type L2Update = {
+            type: 'l2update'
+            product_id: string
+            changes: [string, string, string][] // [side, price, new size]
```
blair: Since you're here, you could add a Side type and change all 'side: string' to 'side: Side'.

export declare type Side = 'buy' | 'sell';

index.d.ts
```diff
@@ -205,20 +208,20 @@ declare module 'gdax' {
         placeOrder(params: OrderParams, callback: callback<OrderResult>): void;
         placeOrder(params: OrderParams): Promise<OrderResult>;
 
-        cancelOrder(orderID: any, callback: callback<string>): void;
-        cancelOrder(orderID: any): Promise<string>;
+        cancelOrder(orderID: string, callback: callback<string | ErrorMessage>): void;
+        cancelOrder(orderID: string): Promise<string | ErrorMessage>;
```
fb55: The promises don't actually resolve to errors

fb55:
This looks great! Just some nits, otherwise should be good to merge :)

index.d.ts
blair: It looks like my previous commit fixed the promise part but missed the orderID part. A rebase against master should point these out and fix them.

fb55:
Thanks a lot @eddiew and @blair!

PR #293, opened 2018-04-01
Add types request and declare http error interface

blair:


fb55:
I would like to replace request with superagent. I will keep this open as a reminder to add a type for this.

blair:
Out of curiosity, why superagent and not request? I did some Googling but didn't find anything documenting why one over the other.

Moving to superagent would make it consistent with gdax-trading-toolkit, which is good.

fb55:
Superagent also works in the browser out of the box. Plus, we use it heavily in most of our JS codebases, so this standardizes things.

blair:
Can you merge this PR and it'll be a hint to switch over to @types/superagent when request is replaced?

fb55:
> Can you merge this PR and it'll be a hint to switch over to when request is replaced?

That is a great reason to do so. Thanks a lot!

PR #294, opened 2018-04-01
index.d.ts: cancelOrder() returns a list of order IDs.

blair:
DELETE /orders/${id} always returns a list of order IDs, even when
deleting a single order.


fb55:
Nice catch!

PR #295, opened 2018-04-01
When a request fails, always set the data property.

blair:
This is for consistency between the different ways an Error instance
is created and populated.


fb55:
Thanks a lot!

PR #302, opened 2018-04-08
Add ticker message definition

codyaray:


codyaray:
Okay, so it seems like this should roughly be an `AggregateTicker` (includes only OHLC type info, except no close price apparently) and a `RawTicker` (which includes both aggregate stats and an individual trade).

What's interesting is that the docs don't correspond to either of these types. They include "trade_id"/"side"/"last_size" like the RawTicker but exclude the open/high/low which are present in both types.

index.d.ts
```diff
@@ -323,6 +354,7 @@ declare module 'gdax' {
         | WebsocketMessage.L2Update
         | WebsocketMessage.Received
         | WebsocketMessage.Match
+        | WebsocketMessage.Ticker
```
fb55: Isn't this undefined?

fb55:
Great work figuring things out, we should add to the docs here.

Reality is actually even more complicated. `open`, `high` and `low` will be ommitted if they haven't been initialized. This should be pretty rare, but it can happen when the box has just been deployed or the stats backend has issues.  Ideally they are marked as optionals, to ensure users ensure their presence.

blair:
@fb55 BTW, do you know what's happening to gdax-trading-toolkit? I've got a large number of PRs and it's becoming hard to manage. I base my PRs off of their master, but there's many conflicts between them if they all are tried to merge in. https://github.com/coinbase/gdax-tt/pulls . I'm thinking of just maintaining my own master and not trying to keep stuff easy to merge back to gdax's master.

codyaray:
pushed another change.

looks like i started adding subscribe/unsubscribe but haven't finished yet. i'm pretty new to typescript and hadn't figured out what the args should look like yet. need to get back to this at some point, but it may be a bit. unless one of you knows what it should look like quickly...

PR #303, opened 2018-04-08
Fix cancelAllOrders compile error.

blair:
Make the args argument required, but make its product_id property
optional, so it's still easy to not set it.

index.d.ts(225,56): error TS1016: A required parameter cannot follow an optional parameter.


fuksi:
@fb55 It seems that this change is not published yet, since the latest version still have the error  https://github.com/coinbase/gdax-node/blob/v0.7.0/index.d.ts#L225

Issue #306, opened 2018-04-13
TypeScript getProductTrades with options fails


wilwade:
TypeScript 2.8.1

```
const opts = { before: 1599086, after: 1599088, limit: 10 };
const trades = await publicClient.getProductTrades('BTC-USD', opts);
```
Results in a ts error: 
```
TS2345: Argument of type '{ before: number; after: number; limit: number; }' is not assignable to parameter of type 'callback<any>'.
  Type '{ before: number; after: number; limit: number; }' provides no match for the signature '(err: any, response: Response, data: any): void'.
```

It appears that ts is matching the first def instead of the second here: https://github.com/coinbase/gdax-node/blob/master/index.d.ts#L165

Perhaps it is the newer version of TypeScript, but I'll throw up a small pr with one fix for it.

fb55:
Fixed in #307

PR #307, opened 2018-04-13
Add pageArgs to the getProductTrades definition

wilwade:
Fix for issue: https://github.com/coinbase/gdax-node/issues/306

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -163,7 +163,10 @@ declare module 'gdax' {
         getProductTicker(productID: string, ): Promise<ProductTicker>;
 
         getProductTrades(productID: string, callback: callback<any>): void;
-        getProductTrades(productID: string, ): Promise<any>;
+        getProductTrades(productID: string): Promise<any>;
+
+        getProductTrades(productID: string, pageArgs: PageArgs, callback: callback<any>): void;
+        getProductTrades(productID: string, pageArgs: PageArgs): Promise<any>;
 
         getProductTradeStream(productID: string, tradesFrom: number, tradesTo: any, callback: callback<any>): void;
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -162,8 +162,8 @@ declare module 'gdax' {
         getProductTicker(productID: string, callback: callback<ProductTicker>): void;
         getProductTicker(productID: string, ): Promise<ProductTicker>;
 
-        getProductTrades(productID: string, callback: callback<any>): void;
-        getProductTrades(productID: string, ): Promise<any>;
+        getProductTrades(productID: string, pageArgs: PageArgs, callback: callback<any>): void;
```
fb55: Can this be marked as optional?

wilwade: It looks like it needs a second set of defs. I'll get that fixed up

blair: We need both lines, since one returns a Promise and the other returns void and has a callback arg. All the gdax-node functions take a callback arg and use it if provided. If there is no callback then the functions return a Promise.

This works, however, the optional isn't consistent with the rest of the file. Maybe make options optional for all the other ones that return a Promise.

```
        getProductTrades(productID: string, options: PageArgs, callback: callback<any>): void;
        getProductTrades(productID: string, options?: PageArgs): Promise<any>;
```

wilwade: @blair I pushed up a match to how the rest of the file does it. There are others on the PublicClient that are still missing types for the options. If you would like, I can in a different pr go back through and add more types and switch the promises over to use the `?` syntax.

blair: The new commit LGTM. How much work you want to put into updating all the calls up to you, but this looks fine for me.

fb55:
Thanks @wilwade and @blair!

Issue #310, opened 2018-04-20
BCH Missing from CurrencyType


wilwade:
Just a quick note that BCH is missing from the CurrencyType. PR coming.

PR #311, opened 2018-04-20
Add BCH to CurrencyType

wilwade:
Quick fix to add BCH to the CurrencyType (Issue: https://github.com/coinbase/gdax-node/issues/310)


fb55:
Nice, thanks!

PR #314, opened 2018-04-24
Add depositPayment method to support /deposits/payment-method

brendanconnolly:
The current authenticated client does not support depositing funds from a payment method.  

The deposit command wraps the /deposits/coinbase-account for moving funds between coinbase accounts. In order to initiate a transfer a call to /deposits/coinbase-account is needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -464,6 +464,18 @@ const withdrawAddressParams = {
 authedClient.withdrawCrypto(withdrawAddressParams, callback);
 ```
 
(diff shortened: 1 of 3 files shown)


fb55:
Could you add documentation and tests?

brendanconnolly:
sure will do

fb55:
Sorry for the delay, @brendanconnolly, and thanks a lot for this patch!

PR #316, opened 2018-05-14
CircleCI Build Matrix

### LINKED ISSUE #249: Improve Build Matrix
I'm extracting this from my comment on https://github.com/coinbase/gdax-node/issues/56#issuecomment-360166288

At a minimum, I'd like to see the build matrix run the test suite against:

* Existing LTS version of node (currently v6 _"Boron"_)
* Newer LTS version of node (currently v8 _"Carbon"_)
* Latest _unstable_ development version of node (currently v9), but potentially allowing the build to fail on this version without holding anything up.

There's also a potential for testing against different permutations of our dependencies, but that's less of an issue for general node package management and it only really has benefits if you're also using a tool like `yarn`.

I'd also like to see the testing tasks improved a little. I know we use `lint-staged` (I'm not a fan, btw, [because it breaks advanced git workflows](https://github.com/okonet/lint-staged/issues/62) ... that's a separate topic), but the test suite should also perform a sanity run of `eslint` and `prettier` to double check that pull-requests weren't submitted with the person's precommit hooks disabled. 

For example, the CI tests should look something like:

```
npm run eslint
npm run prettier
npm test
```

----

> While possible with parallel jobs, Circle isn't the best tool for this. Circle is good tool for private projects and applications, but Travis is far better tool for build matrices and libraries -- especially for open source projects.

@fb55 Thoughts here?

rmm5t:
## What does it do?

- Adds CircleCI build matrix (with node v6, v8, v10)
- Switched to CircleCI version 2
- Also runs a `lint` and `prettier` sanity test prior to running the standard test suite (for people who might disable the `precommit` hooks while using `lint-staged`).

- Also fixes existing prettier errors
- Also fixes build on Node 6 (timing and `ECONNRESET` issues) on `OrderbookSync` and `WebsocketClient`

- `prettier` is only run against `*.js` files, because different npm versions mess with package.json & package-lock.json differently (causing the build to unnecessarily fail on different versions).

## Related Issues

- Closes #249

## Screenshots

gdax-node 2018-05-14 12-20-53

/cc @fb55

### DIFF
--- a/tests/orderbook_sync.spec.js
+++ b/tests/orderbook_sync.spec.js
@@ -98,13 +98,15 @@ suite('OrderbookSync', () => {
           test: true,
           product_id: 'BTC-USD',
         });
-        server.close();
-        done();
       });
     });
 
     server.on('connection', socket => {
       socket.send(JSON.stringify({ test: true, product_id: 'BTC-USD' }));
+      socket.on('message', () => {
+        server.close();
+        done();
+      });
     });
   });
 
@@ -128,13 +130,15 @@ suite('OrderbookSync', () => {
           test: true,
           product_id: 'BTC-USD',
         });
-        server.close();
-        done();
       });
     });
 
     server.on('connection', socket => {
       socket.send(JSON.stringify({ test: true, product_id: 'BTC-USD' }));
+      socket.on('message', () => {
+        server.close();
+        done();
+      });
     });
   });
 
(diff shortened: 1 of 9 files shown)


tests/orderbook_sync.spec.js
```diff
@@ -245,14 +250,16 @@ suite('OrderbookSync', () => {
...
-         assert.equal(productID, 'BTC-USD');
-         server.close();
-         done();
+        assert.equal(productID, 'BTC-USD');
       });
     });
 
     server.on('connection', socket => {
       socket.send(JSON.stringify({ product_id: 'BTC-USD' }));
+      socket.on('message', () => {
```
fb55: Nice find :)

fb55:
Looks great, thanks a lot!

PR #318, opened 2018-05-15
Point Changelog at GitHub Releases

### LINKED ISSUE #317: Better "Changelog" Management
The current `CHANGELOG.md` is nearly empty. 

Suggestion: Let's switch to a more formal history of changes using GitHub Releases.  

Action Item: I'm going to go back and check git history for different releases and ensure that each npm release has an associated git tag and also has an associated set of documentation related to that tag/release (via GitHub Releases).

If anyone gets to this before me, a list of npm released versions, associated git commit shas, and relevants changes since the prior release would be helpful.

rmm5t:
## What does it do?

- Points Changelog at [GitHub Releases](https://github.com/coinbase/gdax-node/releases)

**More importantly**:
- I also backdated some git tags and documented the past few releases to NPM for everything newer than v0.5.0.
    **https://github.com/coinbase/gdax-node/releases**

## Related Issues

- Closes #317 

/cc @fb55

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,1 @@
-# Change Log
-
-## Unreleased
-
-### Changed
-
-* Modify `lib/orderbook_sync` to be initialized with an array of product IDs instead of one product ID, and have it keep track of multiple books for these products in `orderbookSync.books`. `orderbookSync.book` is removed.
+**See [GitHub Releases](https://github.com/coinbase/gdax-node/releases)**


fb55:
This looks great! Was this done manually, or is this something we could automate in the future?

rmm5t:
> Was this done manually, or is this something we could automate in the future?

For this, it was done manually, but heavily based on the git commit history (using some throw-away emacs macros).

PR #319, opened 2018-05-30
Fix Typescript types for getProductTradeStream

kenfehling:


fb55:
Thanks a lot @kenfehling!

PR #322, opened 2018-06-06
TypeScript: Add options for clients

kenfehling:


fb55:
Thanks @kenfehling!

For some reason Prettier fails on Node@6 — merged this anyway as that's out of scope of this PR. (cc @rmm5t)

kenfehling:
Thanks, yeah I noticed that. I also think the same thing happened with my other pull request: https://github.com/coinbase/gdax-node/pull/319

PR #323, opened 2018-06-13
Fix build matrix

rmm5t:
## What does it do?

- Upgrades dev dependencies
- Fixes a websocket test
- Removes README.md from prettier purview -- Different combinations of node/npm/prettier versions  would cause unnecessary build failures and unnecessary markdown format thrashing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
   },
   "scripts": {
     "lint": "eslint --ext .js ./",
-    "prettier": "prettier -l README.md *.js **/*.js",
+    "prettier": "prettier -l *.js **/*.js",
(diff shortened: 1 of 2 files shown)


fb55:
Dope, thanks :)

rmm5t:
@fb55  🤔 This ended up failing again once it hit the `master` branch. I think we're running into versioning issues with prettier, where the default syntax rules change each time prettier is updated. I think the solution might be to freeze prettier to much more specific version. I'll look into submitting another PR related to this.

rmm5t:
Fixed in #324

PR #324, opened 2018-06-14
Fix build and test dependencies

rmm5t:
Redo of #323.

It was necessary to better freeze version of `prettier` so that the build can run more consistently across node and npm versions. 

Also updated the package-lock.json based on Node 8/npm 5.


rmm5t:
😅 Phew, finally.

gdax-node the official node js library for the gdax api 2018-06-14 11-08-24

PR #346, opened 2018-10-27
Rename gdax to coinbase pro

blair:
A bunch of small commits to switch from GDAX to Coinbase Pro. I chose Coinbase Pro since it looks like it would be more popular than Coinbase Prime for users of this module.


fb55:
Thanks so much for this!

PR #371, opened 2019-07-06
#370 Add getPaymentMethods to the AuthenticatedClient class types def…

### LINKED ISSUE #370: Property 'getPaymentMethods' does not exist on type 'AuthenticatedClient'
issue: If your using typescript, getPaymentMethods does not exist on the authenticated client, yet should be.

Steps to reproduce:
Simply create a typescript node project, connect to an authenticated client and call the method. You will get an error at compile time because the method is not defined on the Authenticated Client.

theRealScoobaSteve:
Fixes #370, Did not see a point in adding a test for this one.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -218,6 +218,9 @@ declare module 'coinbase-pro' {
     getCoinbaseAccounts(callback: callback<CoinbaseAccount[]>): void;
     getCoinbaseAccounts(): Promise<CoinbaseAccount[]>;
 
+    getPaymentMethods(callback: callback<CoinbaseAccount[]>): void;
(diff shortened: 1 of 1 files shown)


jfl209:
@theRealScoobaSteve I believe getPaymentMethods needs to return an array of PaymentMethods rather than a CoinbaseAccounts?

https://github.com/coinbase/coinbase-pro-node/pull/375

theRealScoobaSteve:
> @theRealScoobaSteve I believe getPaymentMethods needs to return an array of PaymentMethods rather than a CoinbaseAccounts?
> 
> #375
I would say you are correct sir, will fix

drewrothstein:

PR #375, opened 2019-07-19
Define and return PaymentMethod type, and getPaymentMethods method

### LINKED ISSUE #370: Property 'getPaymentMethods' does not exist on type 'AuthenticatedClient'

jfl209:
Fixes #370

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -124,6 +124,38 @@ declare module 'coinbase-pro' {
     active: boolean;
   };
 
+  export type PaymentMethod = {
+    id: string,
+    type: string,
+    name: string,
+    currency: CurrencyType,
+    primary_buy: boolean,
+    primary_sell: boolean,
+    allow_buy: boolean,
+    allow_sell: boolean,
+    allow_deposit: boolean,
+    allow_withdraw: boolean
+    limits: PaymentMethodLimits
+  };
+
+  export type PaymentMethodLimit = {
+    period_id_days: number,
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -124,6 +124,38 @@ declare module 'coinbase-pro' {
...
+    primary_buy: boolean,
+    primary_sell: boolean,
+    allow_buy: boolean,
+    allow_sell: boolean,
+    allow_deposit: boolean,
+    allow_withdraw: boolean
+    limits: PaymentMethodLimits
+  };
+
+  export type PaymentMethodLimit = {
```
blair, line 132: Can you sort the definitions by use, so PaymentMethodLimitAmount first, then PaymentMethodLimit, PaymentMethodLimits and then PaymentMethod. That seems to be the style used in the rest of the file.

jfl209: 👍

jfl209:
It appears other methods are missing from the typescript declarations:

```
depositPayment
depositCrypto
withdrawPayment
createReport
getReportStatus
```

I'm going to add them to this pull request they are effectively the same (cause) as #370

jfl209:
@blair How do I go about getting a merge?

drewrothstein:

